PR #16, opened 2018-09-26
Add ARD option for non-linear kernel

marpulli:
Add ARD option for creating non-linear multi-fidelity model kernels.

I'm not sure whether we should inherit ARD name from GPy, or choose something that is a bit clearer?


marpulli:
I added a test and decided to keep the ARD name as it seems common across many other GP packages

Issue #2, opened 2018-09-19
Add discrete variables in the class space


javiergonzalezh:

marpulli:
Addressed by PR #17

Issue #15, opened 2018-09-25
Implement context variables


marpulli:
This will allow us to fix certain parameters during the optimization

apaleyes:
A bit more context, sorry for the word play.

Context should probably be a  property of the parameter space, so that you could write code like that:

```
space.fix_context(var1=1)
run_stuff()
space.reset_context()
```

javiergonzalezh:
Question: are we considering that we can set a variable to a value even if is not defined as a context in the initial domain?

marpulli:
@javiergonzalezh what's the use case for that?

marpulli:
Here is what I propose for using the context in the optimization:

Change the signature of the candidate point calculator method to: `CandidatePointCalculator.compute_next_points(self, loop_state: LoopState, context: dict=None) -> np.ndarray`

The `OuterLoop` calls `ParameterSpace.get_context_manager()` to get a dictionary that descrives the context variables then passes it in to the point calculators. This dictionary will be converted to a GPyOpt context manager inside `AcquisitionOptimizer.optimize()` and we keep using the GPyOpt optimizers.

@apaleyes  What do you think?

marpulli:
Done with #45

PR #17, opened 2018-09-27
Mf classes

marpulli:
Addresses issue #2 and adds classes for multi-source decision making loops.

Commits can be squashed when merging.


marpulli:
I've managed to do something bad when merging the old repo which had this changes with the github version, I'll fix it.

emukit/core/loop/candidate_point_calculators.py
```diff
@@ -37,3 +38,51 @@ def compute_next_points(self, loop_state: LoopState) -> np.ndarray:
...
+        self.space = space
+        self.source_parameter = self._get_source_parameter()
+        self.n_sources = np.array(self.source_parameter.domain).size
+
+    def _get_source_parameter(self) -> InformationSourceParameter:
+        """
+        :return: The parameter containing the index of the information source
+        """
+        source_parameter = [param for param in self.space.parameters if isinstance(param, InformationSourceParameter)]
+        if len(source_parameter) == 0:
```
apaleyes: Maybe we should move this logic to parameter space initialization? I find it hard to imagine a valid use case of multiple information source parameters.

emukit/core/loop/user_function.py
```diff
@@ -47,3 +47,50 @@ def evaluate(self, inputs: np.ndarray) -> List[UserFunctionResult]:
...
+        :param inputs: A list of inputs to evaluate the function at with information source index appended as last column
+        :return: A list of function outputs
+        """
+
+        if inputs.ndim != 2:
+            raise ValueError("User function should receive 2d array as an input, actual input dimensionality is {}".format(inputs.ndim))
+
+        n_sources = len(self.f)
+
+        # Run function for inputs at the first information source
```
apaleyes: Can you explain why the code for the first source should be separate from the rest (sorry if I have already asked)? Also, if there is a valid reason, there is some code duplication here, might be worth writing a helper function

emukit/core/optimization/acquisition_optimizer.py
```diff
@@ -10,19 +10,22 @@ def __init__(self, space: ParameterSpace, **kwargs) -> None:
         self.gpyopt_space = space.convert_to_gpyopt_design_space()
         self.gpyopt_acquisition_optimizer = GPyOpt.optimization.AcquisitionOptimizer(self.gpyopt_space, **kwargs)
 
-    def optimize(self, acquisition: Acquisition):
+    def optimize(self, acquisition: Acquisition, context: dict = None):
```
apaleyes: No need to do anything for this PR, but we should take care of #15 and refactor this. I've marked it with 1.0

emukit/core/optimization/acquisition_optimizer.py
```diff
@@ -10,19 +10,22 @@ def __init__(self, space: ParameterSpace, **kwargs) -> None:
...
+
         # Take negative of acquisition function because they are to be maximised and the optimizers minimise
         f = lambda x: -acquisition.evaluate(x)
 
         def f_df(x):
             f_value, df_value = acquisition.evaluate_with_gradients(x)
             return -f_value, -df_value
 
-        if acquisition.has_gradients():
+        if acquisition.has_gradients:
```
apaleyes: Weird, I thought that was fixed.

emukit/core/parameter_space.py
```diff
@@ -22,14 +23,18 @@ def convert_to_gpyopt_design_space(self):
...
-            gpyopt_param = {'name': parameter.name, 'type': 'continuous', 'domain': (parameter.min, parameter.max),
-                            'dimensionality': 1}
+            if isinstance(parameter, ContinuousParameter):
+                gpyopt_param = {'name': parameter.name, 'type': 'continuous', 'domain': (parameter.min, parameter.max),
+                                'dimensionality': 1}
+            elif isinstance(parameter, DiscreteParameter):
+                gpyopt_param = {'name': parameter.name, 'type': 'discrete', 'domain': parameter.domain,
+                                'dimensionality': 1}
+            else:
+                raise NotImplementedError("Nothing except continuous or ordinal is supported right now")
```
apaleyes: ordinal -> discrete

emukit/core/parameter_space.py
```diff
@@ -51,3 +56,9 @@ def check_points_in_domain(self, x: np.ndarray) -> np.ndarray:
             # Set in_domain to be False if this parameter or any previous parameter is out of domain
             in_domain = np.all([in_domain, param_in_domain], axis=0)
         return in_domain
+
+    def append(self, new_parameter):
```
apaleyes: This is a bit scary - changing space on the fly. I'd like to think of parameter space as an immutable thing once defined.

Is it used? Can we avoid using it?

tests/emukit/core/test_parameters.py
```diff
@@ -0,0 +1,8 @@
+import numpy as np
+from emukit.core import DiscreteParameter
+
+
+def test_ordinal_parameter():
```
apaleyes: ordinal -> discrete

emukit/core/loop/candidate_point_calculators.py
marpulli: So we would have a `get_information_source_parameter()` method on `ParameterSpace`? I can do that

emukit/core/__init__.py
```diff
@@ -1,3 +1,4 @@
 # Implement your code here.
```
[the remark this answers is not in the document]
marpulli: sure...

emukit/core/loop/user_function.py
marpulli: I'll refactor, it was to avoid having to know the number of outputs before starting the loop but I'll do it a different way.

emukit/core/parameter_space.py
marpulli: Changed

emukit/core/parameter_space.py
marpulli: Removed

tests/emukit/core/test_parameters.py
marpulli: Changed

emukit/core/loop/candidate_point_calculators.py
apaleyes: Sorry, I meant validation logic, not that method. I don't think we need a method like this

emukit/core/discrete_parameter.py
```diff
@@ -0,0 +1,32 @@
+from typing import Iterable
+
+
+class DiscreteParameter(object):
+    """
+    A parameter that takes a discrete set of values where the order is important
```
mmahsereci: Is only the order important or also the spacing? Let's say we have {1, 2, 3}. Is this equivalent to {1, 2, 4}?

emukit/core/discrete_parameter.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    def __init__(self, name: str, domain: Iterable):
+        """
+
```
mmahsereci: delete blank line?

emukit/core/discrete_parameter.py
```diff
@@ -0,0 +1,32 @@
...
+        """
+        self.name = name
+        self.domain = domain
+
+    def check_in_domain(self, x: Iterable) -> bool:
+        """
+        Checks if the points in x are in the set of allowed values
+
+        :param x: 1d numpy array of points to check
+        :return: A 1d numpy array which contains a boolean indicating whether each point is in domain
```
mmahsereci: type annotations in function call and docstring do not match

emukit/core/loop/candidate_point_calculators.py
```diff
@@ -37,3 +38,49 @@ def compute_next_points(self, loop_state: LoopState) -> np.ndarray:
         """
         x, _ = self.acquisition_optimizer.optimize(self.acquisition)
         return np.atleast_2d(x)
+
+
+class MultiSourceSequential(CandidatePointCalculator):
+    """
+    This candidate point calculator optimizes the acquisition function separately for each information source,
```
mmahsereci: what does `separately' mean in that context? sequentially?

emukit/core/parameter_space.py
```diff
@@ -4,12 +4,18 @@
 import GPyOpt
 
 from . import ContinuousParameter
+from .discrete_parameter import DiscreteParameter, InformationSourceParameter
 
 
 class ParameterSpace(object):
```
mmahsereci: this class is lacking a dockstring.

emukit/core/parameter_space.py
```diff
@@ -4,12 +4,18 @@
...
     def __init__(self, parameters: List):
```
mmahsereci: return type annotation missing

emukit/core/discrete_parameter.py
marpulli: Yep, spacing is important, I'll add this

emukit/core/discrete_parameter.py
marpulli: will do

emukit/core/discrete_parameter.py
marpulli: Good point, I'll change the docstring, the type annotation is correct.

emukit/core/loop/candidate_point_calculators.py
marpulli: It means that there are separate optimizations for each information source parameters and we don't jointly optimize the input location and information source parameter.

emukit/core/parameter_space.py
marpulli: It is but I didn't add this class so I'll leave this for a different PR

emukit/core/parameter_space.py
marpulli: I think we can skip return type annotations for __init__ methods as they don't return anything

emukit/core/loop/candidate_point_calculators.py
mmahsereci: Is this an approximation? I guess no? From the docstring I would not know and would need to ask a developer or go through the code. Basically what I am asking is if `separately' implies that you optimize input location per source sequentially or not, and if yes, if that is an approximation to the solution of the optimization problem. Let's make this more clear.

emukit/core/parameter_space.py
mmahsereci: makes sense.

emukit/core/parameter_space.py
mmahsereci: OK, that is fine with me, but let's do it consistently then. 

I personally prefer uniform consistency over implications though, and adding a -> None once in a while does not add too much overhead and leaves the reader not as puzzled as of why something  was done/left out.

emukit/core/loop/candidate_point_calculators.py
marpulli: I'll rephrase and move the the `compute_next_points()` method as there is too much detail for class docstring

emukit/core/parameter_space.py
marpulli: Sure, let's agree on one and stick with it. I'll leave for a different issue/PR as I didn't touch this init method in this PR and there are quite a few that are inconsistent across the code. Feel free to open an issue!

apaleyes:
When merging, don't forget to update the base branch, as github suggests

Issue #12, opened 2018-09-24
Create the documentation


apaleyes:
Emukit should have documentation up and running on some doc hosting, e.g. RTD. Here is the scope:

* Generate docs for API
* Add notebooks to the docs
* Additional docs on architecture of the library
* Index page for all this stuff above
* Host the documentation

apaleyes:
The only thing left at the moment is to flip RTD default branch once we merge develop to master. Right now "latest" in RTD points to develop.

PR #21, opened 2018-10-06
First version of documentation for Emukit

apaleyes:
*Issue #, if available:* #12 

*Description of changes:*
Basic Sphinx-based documentation. Includes short intro, installation page, tutorials page with notebooks, and API reference.

API is currently generated with sphinx-apidoc, which creates a lot of garbage, but I think that should do for now, given how often we expect our module structure to change in coming weeks, because it is quick and fully automated. As the package structure stabilizes we should come up with a nicer solution.


marpulli:
I built the docs and they look good. Do we need to add the individual .rst files every time we add new modules/classes? Is there a way to auto-generate them when building the docs?

apaleyes:
@marpulli eventually this can become more manual, but at the moment all api docs are automatically generated with sphinx-apidoc. I'll update this PR with a file that lists commands to run

mmahsereci:
thanks for adding the README.md

PR #24, opened 2018-10-08
Fix return type

marpulli:
Changed return type of `create_loop_state` from `None` to `LoopState`. Moved function to end of file so `LoopState` class is already defined when using type annotation.

### DIFF
--- a/emukit/core/loop/loop_state.py
+++ b/emukit/core/loop/loop_state.py
@@ -4,22 +4,6 @@ import numpy as np
 
 from .user_function_result import UserFunctionResult
 
-def create_loop_state(x_init: np.ndarray, y_init: np.ndarray) -> None:
-    """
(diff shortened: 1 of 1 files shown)


emukit/core/loop/loop_state.py
```diff
@@ -4,22 +4,6 @@
 
 from .user_function_result import UserFunctionResult
 
-def create_loop_state(x_init: np.ndarray, y_init: np.ndarray) -> None:
```
apaleyes: Why did it change place? Not that it matters much, but makes harder to spot what has actually changed

marpulli: The LoopState class needs to be defined for me to use the type annotation without quotes around the class name.

apaleyes: Ah, right, makes sense

PR #34, opened 2018-10-12
model based sensitivity using Monte Carlo

javiergonzalezh:
*Description of changes:*
Monte Carlo based sensitivity analysis that uses the predictive mean of an emulator as the target function. 

All tests pass, including the new one added for this feature.


emukit/sensitivity/monte_carlo/monte_carlo_sensitivity.py
```diff
@@ -99,3 +99,22 @@ def compute_effects(self, main_sample: np.ndarray=None, fixing_sample: np.ndarra
 
             var_index += 1
         return main_effects, total_effects, total_variance
+
+
+
+class ModelBasedMonteCarloSensitivity(MonteCarloSensitivity):
+    '''
+    Class to compute the sensitivity coefficients of a Gaussian process  This class wraps the model and calls the mean
```
marpulli: This doesn't seem to be GP specific but the comment makes it seem that way.

tests/emukit/sensitivity/test_emukit_sensitivity.py
```diff
@@ -1,12 +1,44 @@
...
 from emukit.sensitivity.test_functions import Ishigami
+from emukit.core.interfaces import IModel
 
 
+class MockModel(IModel):
+    pass
+
+@pytest.fixture
+def model():
+    model = mock.create_autospec(MockModel)
```
marpulli: Can you just mock IModel directly?

tests/emukit/sensitivity/test_emukit_sensitivity.py
```diff
@@ -1,12 +1,44 @@
...
+
+def test_model_based_montecarlo_sensitivity(model):
+
+    space = ParameterSpace([ContinuousParameter('x1', -np.pi, np.pi),
+                            ContinuousParameter('x2', -np.pi, np.pi),
+                            ContinuousParameter('x3', -np.pi, np.pi)])
+
+    num_mc = 1000
+    np.random.seed(0)
+    senstivity_ishigami = ModelBasedMonteCarloSensitivity(model, space)
```
marpulli: Why ishigami? I don't see the Ishigami function used

tests/emukit/sensitivity/test_emukit_sensitivity.py
```diff
@@ -1,12 +1,44 @@
...
+    senstivity_ishigami.generate_samples(1)
+
+    main_sample = np.array([[ 3.10732573, -1.25504469, -2.66820221], [-1.32105416, -1.45224686, -2.06642419],[ 2.41144646, -1.98949844,  1.66646315]])
+    fixing_sample = np.array([[-2.93034587, -2.62148462,  1.71694805], [-2.70120457,  2.60061313,  3.00826754], [-2.56283615,  2.53817347, -1.00496868]])
+
+    main_effects, total_effects, total_variance = senstivity_ishigami.compute_effects(main_sample = main_sample,fixing_sample=fixing_sample,num_monte_carlo_points = num_mc)
+
+    keys = space.parameter_names
+
+    assert(all(k in main_effects for k in keys))
```
marpulli: As the model gives constant predictions, can we test that the sensitivity indices are what are expected?

emukit/sensitivity/monte_carlo/monte_carlo_sensitivity.py
javiergonzalezh: You are totally right, this class is general, I'll change that.

tests/emukit/sensitivity/test_emukit_sensitivity.py
javiergonzalezh: updated.

mmahsereci:
This is not part of the commit but some of the formulas do not render properly in the notebook. There might be a linebreak missing or so.

tests/emukit/sensitivity/test_emukit_sensitivity.py
javiergonzalezh: yap, done.

tests/emukit/sensitivity/test_emukit_sensitivity.py
javiergonzalezh: yap, done.

javiergonzalezh:
Is funny with the formulas because they render locally jupyter. I'll go over all the notebooks anyway later so I'll look into this in another PR.

mmahsereci:
Yes, I know. The rendering issues seem to appear here and on the doc page for some reason.

emukit/sensitivity/monte_carlo/monte_carlo_sensitivity.py
```diff
@@ -99,3 +99,22 @@ def compute_effects(self, main_sample: np.ndarray=None, fixing_sample: np.ndarra
...
+    '''
+
+    def __init__(self, model: IModel, input_domain: ParameterSpace)-> None:
+        '''
+        :param model: model wrapper with the interface IModel.
+        :param input_domain: space class.
+        '''
+
+        self.model = model
+        self.model_objective = lambda x : self.model.predict(x)[0].flatten()
```
apaleyes: Would it be possible to get rid of using `flatten()` in this case? Generally using things like `flatten` or `atleast_2d` is dangerous, as it does not require people to be strict about the dimensions of the data. And that may lead to fucky behavior sometimes. So I'd rather avoid it, if possible

emukit/core/loop/outer_loop.py
```diff
@@ -86,3 +86,5 @@ def custom_step(self) -> None:
         The user can insert custom code into the loop by overloading this method
         """
         pass
+
```
apaleyes: This file does not seem to be changed

emukit/sensitivity/monte_carlo/__init__.py
```diff
@@ -1,5 +1,3 @@
 # Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
-
-
-from .monte_carlo_sensitivity import MonteCarloSensitivity
+from .monte_carlo_sensitivity import MonteCarloSensitivity, ModelBasedMonteCarloSensitivity
```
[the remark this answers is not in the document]
javiergonzalezh: Yap, done, I have updated the code and the notebook.

emukit/sensitivity/monte_carlo/monte_carlo_sensitivity.py
javiergonzalezh: It is actually not needed as the function is wrapped again.

emukit/core/loop/outer_loop.py
javiergonzalezh: Nop, just a space was added by mistake.

tests/emukit/sensitivity/test_emukit_sensitivity.py
marpulli: What I meant was that I thought the main effects and total effects should be zero because the model outputs constant predictions. If that is true, we should test it here. If it's not true, don't worry.  In response to this comment did you add the test below? Please discuss with @apaleyes whether we want this kind of test.

PR #38, opened 2018-10-15
fix BNN / RF model and add unittests

aaronkl:
Fix the interface for Bayesian neural networks and random forest and add some unittests


emukit/models/bohamiann.py
```diff
@@ -16,6 +12,36 @@
     """)
 
 
+import torch
+import torch.nn as nn
+def get_default_network(input_dimensionality: int) -> torch.nn.Module:
```
apaleyes: This is very nice indeed, as an example. It shows how something a bit elaborate, based on PyTorch, can be used in Emukit

aaronkl:
hmm ... github is weird. I did all the changes and pushed them to the branch of my emukit fork that is associated to this PR. But for some reasons github doesn't automatically add these commits to this PR. However, if I generate a new PR it automatically points me to this one. Do you have any idea how I can update this PR? Otherwise I would delete it and generate a complete new one

Issue #18, opened 2018-10-01
cost sensitive loop


aaronkl:
So far Emukit only considers the case where one has only one single model for the objective function. However, there are many cases, for example EIperSec, Fabolas, MTBO, .... , where one has two models: one for the objective function and one for the cost of evaluating the objective function.
It would be fairly straightforward to bootstrap from the already existing BayesianOptimizationLoop module to implement these methods. However, a cost sensitive loops seems to be a fundamental problem that might be also interesting for Bayesian quadrature or Experimental Design and I am wondering whether it makes sense to implement a general CostSensitiveOuterLoop class instead?

marpulli:
Good idea, I agree that many different loops could use the cost model. 

We would either need the OuterLoop to accept two ModelUpdater classes, one for each model, or we would need to make a ModelUpdater class that deals with updating both models. We can make use of the acquisition function algebra to make a cost aware acquisition: `cost_aware_acquisition = acquisition_function/cost_acquisition` does this.

@apaleyes @javiergonzalezh Does this sound sensible to you?

javiergonzalezh:
This makes total sense to me. I'd go for it.

marpulli:
One other component we would need is a UserFunction that tells us how much a function evaluation cost, as well as the result of the function evaluation.

apaleyes:
Yup, all that makes sense. Just one thing - we shouldn't create more outer loop classes for specific purposes. This is what the elements of the loop are for, and that is why they are so abstract and loosely coupled - to create new, then mix and match. So to introduce the concept of cost into Emukit we should create specific cost-aware model updaters, acquisition functions and/or user functions as needed.

marpulli:
I just chatted to Andrei about this. We would like to have as few cost specific parts as possible. Here are some ideas:
- We would pass in a list of `ModelUpdater`s to the `OuterLoop`. For the cost model case it would call `update()` on both model updaters. It should be able to deal with either one or a list of them. We would need to be able to tell a `ModelUpdater` whether it should use `Y` or `cost` as the training targets.
- The `UserFunctionWithCost` would change depending on what exactly the cost is, so we might want to leave this to the user for now, but if there are common cases (such as timing the function evaluation), we could provide this functionality.
- We can probably just add a `cost` field to the `UserFunctionResult` with a default value of `None`.
- We need to add `cost` as a property of `LoopState`.

I agree about not needing any new interface for the acquisitions. 

What do you think?

aaronkl:
[...]

PR #39, opened 2018-10-15
cost sensitive active loops

aaronkl:
Addresses issue #18  for cost sensitive active loops. The PR does the following changes:
 - allows to pass multiple models to the outer_loop such that we can have one model for the objective function and one for the cost
- stores the cost of each function evaluation in the LoopState
- defines a new UserFunctionWithCost wrapper to also store the cost associated with function evaluations
- adds ExpectedImprovementPerCost as an example for cost-sensitive acquisition functions 
- add  CostSensitiveBayesianOptimizationLoop to allows for Bayesian optimization methods that take the cost into account
- adapts unit test to the new interface
- fixes an issue with BO with entropy search where the pmin was never updated inside the loop


emukit/bayesian_optimization/acquisitions/expected_improvement_per_cost.py
```diff
@@ -0,0 +1,61 @@
...
+from ...core.acquisition import Acquisition
+from emukit.bayesian_optimization.acquisitions import ExpectedImprovement
+
+
+class ExpectedImprovementPerCost(Acquisition):
+
+    def __init__(self, model: Union[IModel, IDifferentiable], cost_model: Union[IModel, IDifferentiable],
+                 jitter: np.float64 = np.float64(0))-> None:
+        """
+        This acquisition computes for a given input the improvement over the current best observed value in
```
mmahsereci: the information that it is per cost is missing. Also as of now i believe the cost function can be evaluated exactly? if not there is a distinction in between expectation of ei/cost and expectation ei/expectation cost. this class computes the latter in the more general case.

emukit/bayesian_optimization/acquisitions/expected_improvement_per_cost.py
```diff
@@ -0,0 +1,61 @@
...
+        """
+        Computes the Expected Improvement.
+
+        :param x: points where the acquisition is evaluated.
+        """
+        improvement = self.ei.evaluate(x)
+
+        mean, _ = self.cost_model.predict(x)
+
+        return improvement / mean
```
mmahsereci: does the cost model make sure that cost mean is not zero?

emukit/bayesian_optimization/acquisitions/expected_improvement_per_cost.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def evaluate(self, x: np.ndarray) -> np.ndarray:
+        """
+        Computes the Expected Improvement.
+
+        :param x: points where the acquisition is evaluated.
+        """
+        improvement = self.ei.evaluate(x)
+
+        mean, _ = self.cost_model.predict(x)
```
mmahsereci: let's give this thing another name. maybe cost instead if mean?

emukit/bayesian_optimization/acquisitions/expected_improvement_per_cost.py
```diff
@@ -0,0 +1,61 @@
...
+
+        return improvement / mean
+
+    def evaluate_with_gradients(self, x: np.ndarray) -> Tuple:
+        """
+        Computes the Expected Improvement and its derivative.
```
mmahsereci: expected improvement per cost

emukit/bayesian_optimization/acquisitions/expected_improvement_per_cost.py
```diff
@@ -0,0 +1,61 @@
...
+
+        :param x: locations where the evaluation with gradients is done.
+        """
+
+        improvement, dimprovement_dx = self.ei.evaluate_with_gradients(x)
+
+        mean, _ = self.cost_model.predict(x)
```
mmahsereci: let's call it cost here, too

emukit/bayesian_optimization/acquisitions/expected_improvement_per_cost.py
```diff
@@ -0,0 +1,61 @@
...
+
+        dmean_dx, _ = self.model.get_prediction_gradients(x)
```
mmahsereci: more descriptive name, please. is it the gradient of the mean of the gp?

emukit/core/loop/loop_state.py
```diff
@@ -9,6 +9,23 @@
 from .user_function_result import UserFunctionResult
 
 
+def create_loop_state(x_init: np.ndarray, y_init: np.ndarray) -> None:
```
javiergonzalezh: When used in a loop with cost, we may want to save in the loop state the values of the cost at each x. For instance, if the cost model is accounting for the time of running an evaluation of the user objective. I think it makes sense having some LoopStateWithCost to handle that.

emukit/core/loop/outer_loop.py
```diff
@@ -27,15 +27,15 @@ class OuterLoop(object):
        criterion is met.
     2. Emukit only calculates the next points(s) to try and you evaluate your function or perform the experiment.
     """
-    def __init__(self, candidate_point_calculator: CandidatePointCalculator, model_updater: ModelUpdater,
+    def __init__(self, candidate_point_calculator: CandidatePointCalculator, model_updaters: List[ModelUpdater],
                  loop_state: LoopState = None) -> None:
         """
         :param candidate_point_calculator: Finds next points to evaluate by optimizing the acquisition function
-        :param model_updater: Updates the data in the model and the model hyper-parameters when we observe new data
+        :param model_updaters: Updates the data in the model and the model hyper-parameters when we observe new data
```
javiergonzalezh: Isn't it better to call this `models_updater`?

emukit/bayesian_optimization/acquisitions/expected_improvement_per_cost.py
aaronkl: you're right I will make that more clear

emukit/bayesian_optimization/acquisitions/expected_improvement_per_cost.py
aaronkl: good point, to prevent that I usually model the log cost. We could either model the log cost as default or just add a very small value to the cost for numerical stability. What do you think?

emukit/bayesian_optimization/acquisitions/expected_improvement_per_cost.py
aaronkl: agreed

emukit/core/loop/loop_state.py
aaronkl: I am also not happy about this right now. `LoopState` already has a property C which stores the cost of each function value. However, `create_loop_state()` only passes X / Y to the `LoopState`. Should we extent `create_loop_state(x_init , y_init)` to `create_loop_state(x_init , y_init, c_init=None)` ?

emukit/bayesian_optimization/acquisitions/expected_improvement_per_cost.py
marpulli: I like the idea of modelling log cost more because it enforces positivity. We might also need to add a small positive value in case the cost is extremely small.

emukit/core/loop/loop_state.py
marpulli: That plan sounds good to me. Two things: this PR seems to undo a previous PR I made where I fixed the return type for this function. Can we call it `cost` rather than `c` everywhere please?

emukit/core/loop/outer_loop.py
marpulli: I think model updaters is fine because it is two separate model updater objects.

emukit/bayesian_optimization/acquisitions/expected_improvement_per_cost.py
```diff
@@ -0,0 +1,61 @@
+from typing import Tuple, Union
```
apaleyes: I have a proposal for a different implementation of cost. Instead of creating <SomeAcquisitionFunction>PerCost, why don't we make use of acquisitions algebra? Let's define a general CostAcquisition. Then all we need to do in CostSensitiveBOLoop would be just divide, and that's it.

The worry here is that we may end up duplicating a lot of stuff, having cost-aware and cost-unaware versions of our code. So I am trying to think of a good way to keep this duplication at a minimum.

It also has an advantage that this cost acquisition function can be re-used in other packages, such as quadrature or exp design.

Makes sense?

emukit/bayesian_optimization/acquisitions/expected_improvement_per_cost.py
mmahsereci: the transformation should happen inside the cost model though. There might be other transformations which are suitable especially when cost_model is a distribution itself. for example analytic approximations to xi-square distributions are easier. 

I agree that we might need to add a small positive value, possibly even a bound, This is because we might run into trouble with the explorative behavior of the cost-sensitive acquisition if the cost is not strictly bounded away from zero (bounded below by zero might not be enough).

emukit/core/loop/loop_state.py
mmahsereci: A general comment to the name. We seem to call everything "cost" not matter if it is just a re-weighing of the acquisition point because for whatever reason we want to enforce certain points to be evaluated more often and other less. 

And second, a "cost" in the classic sense which influences the budget we already used to determine the stopping of the loop. So we kind of have three different scenarios:

1) a custom-weighing of candidate points, evaluation-cost same for all points
2) an evaluation-cost for each point which also determines the budget
3) a combination of 1) and 2)

Algebraically, 1) and 2) they are the same for the acquisition function, but 3) is not included. Also 1) and 2) will need to be handled differently by the outer loop. Thoughts?

emukit/core/loop/outer_loop.py
mmahsereci: then let's change the tex after the "Update_ the data in the model_s..."

emukit/core/loop/loop_state.py
```diff
@@ -33,32 +34,44 @@ def update(self, results: List[UserFunctionResult]) -> None:
...
+        :return: Function outputs for all function evaluations in a 2d array: number of points by output dimensions.
         """
         return np.array([result.Y for result in self.results])
 
+    @property
+    def cost(self) -> np.ndarray:
+        """
+        :return: Cost for evaluating the function in a 2d array: number of points by cost dimensions.
+        """
+        return np.array([result.cost for result in self.results])
```
apaleyes: Is it guaranteed that every result object has a cost field, and that this field contains something sensible? Even for non-cost-aware loops?

emukit/core/loop/outer_loop.py
```diff
@@ -32,15 +32,19 @@ class OuterLoop(object):
...
         """
         :param candidate_point_calculator: Finds next points to evaluate by optimizing the acquisition function
         :param model_updater: Updates the data in the model and the model hyper-parameters when we observe new data
         :param loop_state: Object that keeps track of the history of the loop. Default: None, resulting in empty initial state
         """
         self.candidate_point_calculator = candidate_point_calculator
-        self.model_updater = model_updater
+
+        if isinstance(model_updater, list):
+            self.model_updater = model_updater
```
apaleyes: feels like this should be called `model_updaters` now, notice `s` at the end

emukit/core/loop/user_function.py
```diff
@@ -98,3 +99,41 @@ def evaluate(self, inputs: np.ndarray) -> List[UserFunctionResult]:
         for x, y in zip(inputs, outputs_array):
             results.append(UserFunctionResult(x, y))
         return results
+
+
+class UserFunctionWithCostWrapper(UserFunction):
```
apaleyes: I don't think we need this wrapper. Adding optional cost to the existing wrapper does not look too tricky

emukit/examples/cost_sensitive_bayesian_optimization_loop.py
```diff
@@ -0,0 +1,58 @@
...
+from ..core.acquisition import Acquisition
+from ..core.interfaces import IModel
+from ..core.loop import CandidatePointCalculator, FixedIntervalUpdater, ModelUpdater, OuterLoop, Sequential
+from ..core.loop.loop_state import create_loop_state
+from ..core.optimization import AcquisitionOptimizer
+from ..core.parameter_space import ParameterSpace
+
+
+class CostSensitiveBayesianOptimizationLoop(OuterLoop):
+    def __init__(self, model_objective: IModel, model_cost: IModel,
```
apaleyes: This _init_ is a clear subject to review as a part of #63 . Just saying

emukit/core/loop/loop_state.py
marpulli: It defaults to None so you would get an array of None's here

emukit/bayesian_optimization/acquisitions/acquisition_per_cost.py
```diff
@@ -0,0 +1,64 @@
...
+
+        :param x: A numpy array of shape (n_inputs x n_input_features) containing input locations at which to evaluate
+                  the cost
+        :return: Tuple of numpy arrays: (cost value, cost gradients)
+        """
+        expected_cost = self.cost_model.predict(x)[0]
+        is_below_min_cost = (expected_cost < self.min_cost).flatten()
+        grad = self.cost_model.get_prediction_gradients(x)[0]
+        grad[is_below_min_cost, :] = 0
+        print(is_below_min_cost)
```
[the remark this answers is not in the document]
marpulli: I've just removed these. It's an array so looks messy in the logs

PR #49, opened 2018-10-24
Update model at the end of the loop

marpulli:
This PR makes sure the model is up to date with all observations at the end of the loop. 

Not sure if we should actually do this so would like other people's opinion. If the model training takes a long time and we are sure that we don't need the model to be up to date we would not want to do this but in most cases it seems a sensible thing to do.

### DIFF
--- a/tests/emukit/core/test_outer_loop.py
+++ b/tests/emukit/core/test_outer_loop.py
@@ -1,7 +1,11 @@
+from typing import Tuple
+
 import mock
 import pytest
 
 import numpy as np
+from emukit.core.interfaces import IModel
+
 from emukit.core.loop import (OuterLoop, ModelUpdater, StoppingCondition, CandidatePointCalculator, UserFunctionResult,
                               LoopState, UserFunction)
 
@@ -31,6 +35,7 @@ def mock_user_function():
 
     return user_function
 
+
 def test_outer_loop(mock_next_point_calculator, mock_updater, mock_user_function):
     """ Example of automatic outer loop """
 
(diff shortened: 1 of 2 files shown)


tests/emukit/core/test_outer_loop.py
```diff
@@ -43,6 +44,16 @@ def test_outer_loop(mock_next_point_calculator, mock_updater, mock_user_function
...
 
+
+def test_outer_loop_model_update(mock_next_point_calculator, mock_updater, mock_user_function):
+    """ Checks the model has the correct number of data points """
+
+    loop = OuterLoop(mock_next_point_calculator, mock_updater)
+    loop.run_loop(mock_user_function, 2)
+
+    # Check update was last called with a loop state with all the collected data points
+    assert mock_updater.update.call_args[0][0].X.shape[0] == 2
```
apaleyes: Can you change this test so that it just checks that model has all the data instead? Tests that verify particular function calls are best avoided, as they test implementation details and not the output, and thus tend to be brittle.

marpulli: Hmmm, I'm not convinced. Then I'd need to instantiate a real model and a model updater instead of mocks. Would I create a dummy model class to do this? Actual models would slow the test down with unnecessary model optimizing. I'd still be testing as many implementation details as I am here, as I'd have to extract the data from the loop state.

apaleyes: Have a mock model and implement update_data to just collect data. Then you can compare it with expected in the end, or just the length of it

marpulli: I've changed the test, could you take a look and see if that's what you meant? I feel like I'm now just testing the implementation of my mocks rather than whether the model updater gets the correct number of data points.

apaleyes: In fact it is, thanks!

Issue #32, opened 2018-10-12
Get optional dependencies straight


apaleyes:
We have a few models with optional dependencies at the moment. They are not tested, and result in empty doc pages. We need to find a good place for them to go, and think about how to manage their dependencies appropriately.

PR #51, opened 2018-10-26
Integration tests setup, updated docs

apaleyes:
*Issue #, if available:* #32 

*Description of changes:*
- Replaced RF model with scikit-learn
- Fixed requirements docs
- Setup integration tests run during the build
- Removed examples from coverage reports. It is not trivial to track, and it is questionable whether we should be tracking coverage of examples as strict as the core library
- As a side effect, updated docs to include recent changes

Testing: made sure tests, integration tests and docs all run locally


marpulli:
Seems like `enum.auto` was introduced in 3.6. We can either not support 3.5 or remove the use of `auto`.

.travis.yml
```diff
@@ -27,13 +27,15 @@ install:
   - pip install --upgrade pip
   - pip install -r requirements/requirements.txt
   - pip install -r requirements/test_requirements.txt
+  - pip install -r requirements/integration_test_requirements.txt
```
marpulli: To stop the linux logs reaching the size limit, maybe making the pip install less verbose would work?
```suggestion
  - pip install -r requirements/integration_test_requirements.txt -q
```

emukit/examples/models/random_forest.py
```diff
@@ -3,64 +3,43 @@
...
-        pip install git+https://github.com/automl/random_forest_run.git
+        pip install scikit-learn
+
+        Refer to http://scikit-learn.org/stable/install.html for further information.
     """)
 
 
 class RandomForest(IModel):
 
     def __init__(self, X_init: np.ndarray, Y_init: np.ndarray, num_trees: int = 30,
```
marpulli: how do you feel about allowing the user to pass in kwargs for the sklearn model? We don't want to expose all options in the sklearn random forest constructor but we are limiting what a user can do here.

apaleyes: Personally, I feel negatively, but I can be persuaded... 

Anyways, the sole point of existence of these models, from my point of view, was to give users something simple to play with out of the box. They will be limiting no doubt, but then they are not intended for a real modelling.

.travis.yml
apaleyes: That's a good idea, thanks. Will certainly give it a try.

apaleyes:
@marpulli I was going to get rid of `auto()`, feels wrong to rule out 3.5 just because of it

emukit/examples/models/random_forest.py
marpulli: But if they can be made ready for "real" modelling with such a small change, shouldn't we do it? I think it would be quite frustrating if I had to re-implement this class just so I can pass in one more parameter to the sklearn random forest constructor.

An alternative might be to have a "SciKitLearnRandomForestWrapper" and simply have a function that creates a random forest with default parameters, passes it in to the wrapper then returns this wrapper to the user. Then if a user wants to create a custom random forest model they can reuse the wrapper.

apaleyes:
"will increase coverage by 8.81%."    yay!

Issue #25, opened 2018-10-08
Add ways to customize outer loop flow


apaleyes:
It should be possible to customize the loop flow without rewriting it. For instance, if a user wants to store some benchmarking information between iterations, they should be able to add it into the existing loop without redefining it.

Right now we have `self.custom_step()` at the end of each iteration, which does the job to some extent, but obviously isn't very flexible and does not cover for many scenarios.

javiergonzalezh:
We have 4 main steps in the grand loop: update model, run policy (collect points), evaluate objective and update the state. Maybe an idea is to have a custom call for each one of these components instead of having just one for the entire loop.

marpulli:
If the only need for this at the minute is to save benchmarking information, should we have an optional `ModelScorer` class as part of the outer loop that does this instead? I'm imagining:

```
class ModelScorer:
    def score_model(self, model)
        stuff here...
```

I'm not convinced about the name yet, I'll keep thinking about that

apaleyes:
No, the idea was to give more flexibility in general. At the moment however benchmarking is the only thing that comes to mind.

Implementation-wise i was thinking of something like that:

       self.loop_starting()
       while not stopping_condition.should_stop(self.loop_state):
            self.loop_iteration_started()

            self.model_updating()
            self.model_updater.update(self.loop_state)
            self.model_updated()

            self.next_points_computing()
            new_x = self.candidate_point_calculator.compute_next_points(self.loop_state)
            self.next_points_computed()

            self.objective_executing()
            results = user_function.evaluate(new_x)
            self.objective_executed()

            self.loop_state_updating()
            self.loop_state.update(results)
            self.loop_state_updated()

            self.loop_iteration_finished()

        self.loop_finished()

And when users have a loop object, they would subscribe to these events, maybe with something as simple as `my_loop.loop_state_updated = my_event_handler()`. And these events may have arguments too, say loop state, maybe more.

Hopefully the idea is clear, although of course this is very verbose - we don't really need that many events. Let me know what you think.

apaleyes:
First step in this direction is implemented by @marpulli  in #53 . We will add more customization to the loop as we find use cases for it.

PR #53, opened 2018-10-30
Add loop custom steps

marpulli:
*Issue #, if available:* #25 

*Description of changes:* I've added custom steps at the start and end of the iteration. This is deliberately fewer than suggested in issue #25 as I'm not sure what the use case for any other would be. 


marpulli:
Nice thread, thanks. This one looks neat, I'll try it out: https://stackoverflow.com/a/2022629

marpulli:
Updated with an `EventHandler` class and also removed the event at iteration start, as I have no use case right now. It should be easy to add if we find a use case later

emukit/core/event_handler.py
```diff
@@ -0,0 +1,11 @@
+class EventHandler(list):
+    """
+    A list of callable objects. Calling an instance of this will cause a call to each item in the list in ascending
```
apaleyes: Let's link the SO answer here. I know SO does not demand it, but attribution does not cost anything and is a nice thing to do.

emukit/core/loop/outer_loop.py
```diff
@@ -39,6 +41,7 @@ def __init__(self, candidate_point_calculator: CandidatePointCalculator, model_u
         self.loop_state = loop_state
         if self.loop_state is None:
             self.loop_state = LoopState([])
+        self.loop_iteration_end_event = EventHandler()
```
apaleyes: Can you start a list in the docstring with events this object exposes? At the moment it will have just one item, but I am sure it will grow

Also I think "loop_" part can be dropped from the name, since it is a part of the loop object

tests/emukit/core/test_outer_loop.py
```diff
@@ -1,13 +1,19 @@
-from typing import Tuple
-
 import mock
 import pytest
+from types import MethodType
```
apaleyes: is this used?

tests/emukit/core/test_outer_loop.py
```diff
@@ -68,48 +76,87 @@ def X(self):
...
 
 def test_condition_invalid_type(mock_next_point_calculator, mock_updater, mock_user_function):
     invalid_n_iter = 10.0
     loop = OuterLoop(mock_next_point_calculator, mock_updater)
 
     with pytest.raises(ValueError):
         loop.run_loop(mock_user_function, invalid_n_iter)
+
+
+def test_custom_step_call():
```
apaleyes: this is no longer called "custom_step". how about "test_iteration_end_event"?

emukit/core/event_handler.py
marpulli: Done

emukit/core/loop/outer_loop.py
marpulli: Done both items

tests/emukit/core/test_outer_loop.py
marpulli: Removed

tests/emukit/core/test_outer_loop.py
marpulli: Renamed

PR #60, opened 2018-11-02
Draft of the vision doc

apaleyes:
This doc describes a high level vision for Emukit. I expect a heated debate over its content!


doc/vision.rst
```diff
@@ -0,0 +1,45 @@
...
+
+Emukit and emulation
+_____________________
+
+Here is Emukit approach towards each of three parts of emulation.
+
+Methods
+^^^^^^^^
+
+This is the main focus of Emukit. Emukit defines a general sctructure of a decision making method, called ``OuterLoop``, and then offers implementations of few such methods: Bayesian optimization, sensitivity analysis, experimental design. All methods in Emukit are model-agnostic.
```
javiergonzalezh: Let's remove sensitivity analysis from this paragraph and add a sentence like this: 'In addition to provide a framework for decision making Emukit provide other tools, like sensitivity analysis, that help to debug and interpret emulators'.

doc/vision.rst
```diff
@@ -0,0 +1,45 @@
...
+
+Generally speaking, Emukit does not provide modelling capabilities, instead expecting users to bring their own models. Because of the variety of modelling frameworks out there, Emukit does not mandate or make any assumptions about a particular modelling technique or a library. Instead it suggests to implement a subset of defined model interfaces required to use a particular method. Nevertheless, there are a few model-related functionalities in Emukit:
+- Example models, which give users something to play with to explore Emukit.
+- Model wrappers, which are designed to help adapting models in particular modelling frameworks to Emukit interfaces.
+- Multi-fidelity models, implemented based on `GPy <https://github.com/SheffieldML/GPy>`_.
+
+Tasks
+^^^^^^
+
+Emukit does not contribute much to this part at the moment. However Emukit team are on lookuout for typical use cases for Emukit, and if a reoccuring pattern emerges, it may become a part of the library.
```
javiergonzalezh: In a way, this tasks correspond to the examples folder. Shall we call that folder tasks and mention here that contributions are welcome?

apaleyes: At least according to this vision, tasks are much closer to business goals you may pursue using emukit as a tool. From this perspective examples folder does not look particularly relevant.

Nevertheless I agree, potentially an example of how to solve a business problem may land in there. That is why the doc says "reoccurring". There is no doubt one can work out examples of how to use emukit, but we shouldn't be eager to make them all part of the library.

All in all, "examples" name translates a right connotation, so I'd like to stick to it for the time being.

doc/vision.rst
mmahsereci: quadrature is also not model agnostic.

apaleyes: @mmahsereci , this is in "modelling framework" sense

apaleyes:
Added a picture. It is fairly high level, and does not mention many details (multi fidelity, wrappers, etc), I feel it explains the vision well

Issue #70, opened 2018-11-07
Updating acquisition function in the loop


marpulli:
Some acquisition functions, such as entropy search, require a recalculation of some parameters after each iteration. This doesn't neatly fit into our `OuterLoop` framework at the minute. We may need to add an `update` method to the acquisition which is called after observing new data

marpulli:
#71 addressed this

PR #71, opened 2018-11-08
Add update_between_iterations method

marpulli:
*Issue #, if available:* #70 

*Description of changes:* Added "update_between_iterations" method on an acquisition function, so that any parameters that can be computed once per outer loop iteration can be cached. This is used with the entropy search acquisition


emukit/core/acquisition.py
```diff
@@ -66,6 +66,12 @@ def evaluate_with_gradients(self, x: np.ndarray) -> Tuple[np.ndarray, np.ndarray
         """
         raise NotImplementedError("Gradients not implemented for this acquisition function")
 
+    def update_between_iterations(self) -> None:
```
mmahsereci: Is it usual to call methods by the way they are used in some application rather than what they do?

apaleyes: I second this comment. "iteration" concept does not exist on acquisition on CandPointCalc level. This needs rephrasing to reflect what this method actually does.

Also, I find it weird this method does not take any inputs. Even though ES case does not need it, I wander if it is an update at all. Perhaps ES case is so special it could be re-implemented slightly to avoid this new API altogether?

marpulli: I'll rename as `update_parameters` and add loop state as input

PR #78, opened 2018-11-09
Allow model updater to update from any field in loop state

marpulli:
Currently the cost model in the cost sensitive loop is trained on "Y" values whereas it should be trained on "cost" values. This PR fixes that. 

### DIFF
--- a/emukit/core/loop/model_updaters.py
+++ b/emukit/core/loop/model_updaters.py
@@ -3,6 +3,7 @@
 
 
 import abc
+from typing import Callable
 
 from . import LoopState
 from ..interfaces import IModel
@@ -10,6 +11,7 @@ from ..interfaces import IModel
 
 import logging
 _log = logging.getLogger(__name__)
+
 
 class ModelUpdater(abc.ABC):
     @abc.abstractmethod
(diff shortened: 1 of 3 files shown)


emukit/core/loop/model_updaters.py
```diff
@@ -24,19 +25,22 @@ def update(self, loop_state: LoopState) -> None:
 
 class FixedIntervalUpdater(ModelUpdater):
     """ Updates hyper-parameters every nth iteration, where n is defined by the user """
-    def __init__(self, model: IModel, interval: int = 1) -> None:
+    def __init__(self, model: IModel, interval: int = 1, target_name='Y') -> None:
```
apaleyes: I suggest a lambda to get training data out of loop state instead of this reflection thingy. Something like `lambda loop_state: loop_state.Y`

marpulli: Done

emukit/core/loop/model_updaters.py
```diff
@@ -24,19 +26,27 @@ def update(self, loop_state: LoopState) -> None:
 
 class FixedIntervalUpdater(ModelUpdater):
     """ Updates hyper-parameters every nth iteration, where n is defined by the user """
-    def __init__(self, model: IModel, interval: int = 1) -> None:
+    def __init__(self, model: IModel, interval: int=1, targets_extractor_fcn: Callable=None) -> None:
         """
         :param model: Emukit emulator model
         :param interval: Number of function evaluations between optimizing model hyper-parameters
+        :param targets_extractor_fcn: A lambda function that takes in loop state and returns the training targets.
```
apaleyes: Does not have to be lambda, right? Any function would do
```suggestion
        :param targets_extractor_fcn: A function that takes in loop state and returns the training targets.
```

PR #89, opened 2018-11-15
Updates to sensitivity

apaleyes:
- Removed numerical tests
- Cleaned up comments

To clarify one change here. ''' -> """ in many docstrings is done because of [PEP257](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings). This is also the style we follow elsewhere in the code base.


apaleyes:
@marpulli done

PR #92, opened 2018-11-16
Find change in entropy rather than new entropy in entropy search acquisition

marpulli:
*Description of changes:* Entropy search currently had a term missing that was the current entropy of the distribution over the minimum. This meant that if we evaluated the acquisition at a point that would not change the distribution over the minimum (e.g. existing training point in noiseless model), the acquisition function did not return zero as expected. This matters if we want to use entropy search per unit cost.

@aaronkl does this look right to you?

### DIFF
--- a/emukit/bayesian_optimization/acquisitions/entropy_search.py
+++ b/emukit/bayesian_optimization/acquisitions/entropy_search.py
@@ -128,6 +128,7 @@ class EntropySearch(Acquisition):
         self.logP, self.dlogPdMu, self.dlogPdSigma, self.dlogPdMudMu = epmgp.joint_min(mu, var, with_derivatives=True)
         self.logP = self.logP[:, np.newaxis]
 
+        # Calculate the entropy of the distribution over the minimum given the current model
         self.p_min_entropy = np.sum(np.multiply(np.exp(self.logP), np.add(self.logP, self.representer_points_log)),
                                     axis=0)
 
--- a/emukit/sensitivity/monte_carlo/monte_carlo_sensitivity.py
+++ b/emukit/sensitivity/monte_carlo/monte_carlo_sensitivity.py
@@ -11,25 +11,27 @@ from ...experimental_design.model_free.random_design import RandomDesign
 from ...core.interfaces.models import IModel
 
 class ModelFreeMonteCarloSensitivity(object):
-    '''
+    """
     Class to do sensitivity analysis of a function. It computes Monte Carlo approximations to
     the Sobol indexes and the total variance components of each input variable of some objective
     of interest.
-    '''
+    """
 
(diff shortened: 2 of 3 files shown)


emukit/bayesian_optimization/acquisitions/entropy_search.py
```diff
@@ -180,10 +184,11 @@ def evaluate(self, x: np.ndarray) -> np.ndarray:
         predicted_logP = np.subtract(predicted_logP, lselP)
 
         # We maximize the information gain
-        dHp = np.sum(np.multiply(np.exp(predicted_logP), np.add(predicted_logP, self.representer_points_log)), axis=0)
+        H_p = np.sum(np.multiply(np.exp(predicted_logP), np.add(predicted_logP, self.representer_points_log)), axis=0)
 
-        dH = np.mean(dHp)
-        return np.array([[dH]])
+        new_entropy = np.mean(H_p)
+        entropy_change = new_entropy - self.p_min_entropy
```
aaronkl: I am not entirely sure but I think the averaging happens after we subtracted self.p_min_entropy, i e.:
`entropy_change = np.mean(H_p - self.p_min_entropy)`

emukit/bayesian_optimization/acquisitions/entropy_search.py
```diff
@@ -127,6 +128,9 @@ def update_pmin(self) -> np.ndarray:
         self.logP, self.dlogPdMu, self.dlogPdSigma, self.dlogPdMudMu = epmgp.joint_min(mu, var, with_derivatives=True)
         self.logP = self.logP[:, np.newaxis]
 
+        self.p_min_entropy = np.sum(np.multiply(np.exp(self.logP), np.add(self.logP, self.representer_points_log)),
```
aaronkl: Could we maybe add a comment here to make it clear that this is the entropy of the current p_min distribution based on our model

emukit/bayesian_optimization/acquisitions/entropy_search.py
marpulli: My understanding was that it's after. Equation (6) here:
http://jmlr.csail.mit.edu/papers/volume13/hennig12a/hennig12a.pdf

Either way, `self.p_min_entropy` is a scalar so the numerical result would be identical.

emukit/bayesian_optimization/acquisitions/entropy_search.py
marpulli: Will do

aaronkl: Ah sorry my bad you're right. It is a scalar. We only have to average the predicted p_min later over all halluzinated values. It should be correct this way

PR #94, opened 2018-11-19
Benchmarking Package

marpulli:
This PR contains the following:
- A `Benchmarker` class that runs different loops with different initial data sets.
- A `BenchmarkPlot` class that plots the results of the benchmarking
- A `Metrics` interface and implementations of some common metrics. These metrics are evaluated on every iteration of every loop and results stored so they can be plotted later.
- A `BenchmarkResult` class which stores the metrics.

Questions:
- Are the metrics more general? Would we want to collect metrics about all loops, not just when we are benchmarking?
- Should a loop return some results? `loop.run_loop()` currently doesn't return anything which is a bit weird. If there were a `LoopResult` object, we could reuse it here.


javiergonzalezh:
This is great Mark. I think that in the loop state we can save other metrics that can be queried by the `Benchmarker`. 

I agree with the point that the loops should return something. This may depends on the loop itself. In BO for instance, a user will expect at least the optimum (and  it value) and the current model). In BQ the value of the integral, etc.

emukit/benchmarking/benchmark_plot.py
```diff
@@ -0,0 +1,132 @@
...
+import numpy as np
+
+from .benchmark_result import BenchmarkResult
+
+
+class BenchmarkPlot:
+    """
+    Creates a plot comparing the results from the different loops used during benchmarking
+    """
+    def __init__(self, benchmark_results: BenchmarkResult, loop_colours=None, loop_line_styles: List=None,
```
apaleyes: What type should `loop_colours` be?

emukit/benchmarking/benchmark_plot.py
```diff
@@ -0,0 +1,132 @@
...
+                 x_axis: str=None, metrics_to_plot: List[str]=None):
+        """
+        :param benchmark_results: The output of a benchmark run
+        :param loop_colours: Colours to use for each loop. Defaults to standard matplotlib colour palette
+        :param loop_line_styles: Line style to use for each loop. Defaults to solid line style for all lines
+        :param x_axis: Which metric to use as the x axis in plots. None means it will be plotted against iteration
```
apaleyes: `x_axis_metric_name` would be more descriptive

emukit/benchmarking/benchmark_plot.py
```diff
@@ -0,0 +1,132 @@
...
+            self.metrics_to_plot = self.benchmark_results.metric_names
+        else:
+            for metric_name in metrics_to_plot:
+                if metric_name not in self.benchmark_results.metric_names:
+                    raise ValueError(metric_name + ' not found in saved metrics from benchmark results.')
+            self.metrics_to_plot = metrics_to_plot
+
+        if (x_axis is not None) and (x_axis in self.metrics_to_plot):
+            if x_axis not in self.metrics_to_plot:
+                raise ValueError('x_axis ' + x_axis + ' is not a valid loop name')
```
apaleyes:
```suggestion
                raise ValueError('x_axis ' + x_axis + ' is not a valid metric name')
```

emukit/benchmarking/benchmark_plot.py
```diff
@@ -0,0 +1,132 @@
+from itertools import cycle
+from typing import List
+
+import matplotlib.pyplot as plt
```
apaleyes: We don't declare an explicit dependency on matplotlib. And neither we should, as emukit can be easily used without plotting. However we probably should create an extra requirement, [similar to GPy](https://github.com/SheffieldML/GPy/blob/devel/setup.py#L167), and maybe call it "benchmarking".

Consequently, I'd suggest we  wrap this import in try/except to handle import error, with similar handling to what we do with model examples at the moment.

emukit/benchmarking/benchmark_result.py
```diff
@@ -0,0 +1,47 @@
...
+        Add results for a specific loop, metric and repeat combination
+
+        :param loop_name: Name of loop
+        :param i_repeat: Index of repeat
+        :param metric_name: Name of metric
+        :param metric_values: Metric values to add
+        """
+        self._results[loop_name][metric_name][i_repeat] = np.array(metric_values).flatten()
+
+    def extract_metric_as_array(self, loop_name: str, metric_name: str):
```
apaleyes:
```suggestion
    def extract_metric_as_array(self, loop_name: str, metric_name: str) -> np.ndarray:
```

emukit/benchmarking/benchmark_result.py
```diff
@@ -0,0 +1,47 @@
...
+
+        self._results = dict()
+        for loop_name in loop_names:
+            self._results[loop_name] = dict()
+            for metric_name in metric_names:
+                self._results[loop_name][metric_name] = []
+                for i in range(n_repeats):
+                    self._results[loop_name][metric_name].append([])
+
+    def add_results(self, loop_name: str, i_repeat: int, metric_name: str, metric_values: list) -> None:
```
apaleyes: The fact that metric_values are not already a numpy array feels slightly weird. Is there any reason for that?

emukit/benchmarking/benchmarker.py
```diff
@@ -0,0 +1,87 @@
...
+    def __init__(self, loops_with_names: List[Tuple[str, Callable]], test_function: Union[Callable, UserFunction],
+                 parameter_space: ParameterSpace, metrics: List[Metric], initial_design: ModelFreeDesignBase=None):
+        """
+        :param loops_with_names: A list of tuples where the first entry is the name of the loop and the second is a
+                                 function that takes in initial x and y training data and returns a loop to be
+                                 benchmarked
+        :param test_function: The function to benchmark the loop against
+        :param parameter_space: Parameter space describing the input domain of the function to be benchmarked against
+        :param metrics: List of metric objects that assess the performance of the loop at every iteration
+        :param initial_design: An object that returns an set of samples in the input domain that are used as the initial
```
apaleyes:
```suggestion
        :param initial_design: An object that returns a set of samples in the input domain that are used as the initial
```

emukit/benchmarking/benchmarker.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def run_benchmark(self, n_initial_data: int=10, n_iterations: int=10, n_repeats: int=10) -> BenchmarkResult:
+        """
+        Runs the benchmarking. For each initial data set, every loop is created and run for the specified number of
+        iterations and the results are collected.
+
+        :param n_initial_data: Number of points in the initial data set
+        :param n_iterations: Number of iterations to run the loop for
+        :param n_repeats: Number of times to run each loop with a different initial data set
+        :return: A numpy array with the resulting loop state
```
apaleyes: Nope, it returns an instance of BenchmarkResult

emukit/core/loop/outer_loop.py
```diff
@@ -73,6 +76,8 @@ def run_loop(self, user_function: Union[UserFunction, Callable], stopping_condit
 
         _log.info("Starting outer loop")
 
+        self.loop_start_event(self, self.loop_state)
```
apaleyes: One more event. Yay! For some reason it makes me excited

emukit/benchmarking/metrics.py
```diff
@@ -0,0 +1,80 @@
...
+        :param loop: Outer loop
+        :param loop_state: Object containing history of the loop that we add results to
+        """
+        y_min = np.min(loop_state.Y, axis=0)
+        # Add to metrics dictionary in loop state
+        _add_value_to_metrics_dict(loop_state, y_min, self.name)
+
+
+class TimeMetric(Metric):
+    def __init__(self, name='time'):
```
apaleyes:
```suggestion
    def __init__(self, name: str='time'):
```

tests/emukit/benchmarking/test_metrics.py
```diff
@@ -0,0 +1,48 @@
+import mock
+import numpy as np
+
+from emukit.benchmarking.metrics import MeanSquaredErrorMetric, MinimumObservedValueMetric
+from emukit.core.interfaces import IModel
+from emukit.core.loop import LoopState, ModelUpdater, OuterLoop
+from emukit.core.loop.loop_state import create_loop_state
+
+
+def test_mean_squared_error_metric():
```
apaleyes: some test, even if it does not measure the actual output, for time metric would be nice

integration_tests/emukit/benchmarking/test_benchmarker.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+
+import emukit.test_functions
+from emukit.benchmarking.benchmarker import Benchmarker
+from emukit.benchmarking.metrics import MinimumObservedValueMetric
+from emukit.core import ContinuousParameter
+from emukit.examples.gp_bayesian_optimization.single_objective_bayesian_optimization import GPBayesianOptimization
```
apaleyes: Given that we will probably hold a lower bar for examples folder, let's not rely on it for unit testing core functionality of the library

emukit/benchmarking/benchmarker.py
```diff
@@ -0,0 +1,87 @@
...
+                for metric_name, metric_value in this_loop.loop_state.metrics.items():
+                    result.add_results(loop_name, j, metric_name, metric_value)
+        return result
+
+    def _create_initial_loop_state(self, n_initial_data):
+        x_init = self.initial_design.get_samples(n_initial_data)
+        results = self.test_function.evaluate(x_init)
+        return LoopState(results)
+
+    def _initialise_results(self):
```
apaleyes: Isn't that obsolete now, with BenchmarkResult?

marpulli:
Addressed most of the comment but I'm goign to work on making the metrics a more integral part of the `OuterLoop`

Issue #67, opened 2018-11-06
PEP8 compliance


marpulli:
We say in our contribution guidelines the code should adhere to PEP8 standards. It currently does not. Let's fix it and then check it stays compliant

PR #96, opened 2018-11-20
PEP8 pt1: configuration, fixed test files

apaleyes:
*Issue #, if available:* #67 


apaleyes:
Besides just pure pep8 compliance, this proved useful for other reasons. Namely I discovered that we were missing tests for some parameter types, and that we could simplify some tests with parametrization.

Issue #63, opened 2018-11-05
Use abstractions appropriately


apaleyes:
Before going 1.0 we should review our use of abstractions. Based on recent conversations I propose this approach:

* On OuterLoop level we should be dealing in ModelUpdater, CandPointCalc and so on.
* On concrete method level we should be dealing in problem specific terms: acquisition, update step size, etc.
* If concrete method does not suit user needs, it should be clear how to create own implementation of the OuterLoop. In fact, users should be encouraged to do that.

There are few places in the code base where we currently don't follow this approach, and if everyone agrees with this proposal, we should identify and fix those places.

apaleyes:
@marpulli @javiergonzalezh @mmahsereci let me know your thoughts!

javiergonzalezh:
I agree with this. One of the things that we need to do is to separate the logic of computing the next points from the logic of optimizing an acquisition that is passed in. In the sequential way of collecting points it make sense having them together but as soon as we start computing batches things can be more complex.

javiergonzalezh:
Closing as Mark sorted this one out already.

apaleyes:
Let's keep it open for now, as we still need to review the loop's APIs to make sure they are consistent.

PR #107, opened 2018-11-26
More consistent use of abstractions

marpulli:
*Issue #, if available:* #63 

*Description of changes:*
- Move cost sensitive loop to BO package. It seems to fit better there
- Remove X/Y/Cost inputs from cost sensitive loop inputs. The current implementation relies on Xs to be the same for the cost and objective model which I now check for.
- Removed model updaters and candidiate point calculator from cost sensitive loop
- Removed model updater from BO loop
- Re-order inputs of BO and cost sensitive BO loops to match experimental design
- Update tests
- Update notebook for these changes + some previous changes to APIs


marpulli:
@apaleyes My local git recognises the file was moved rather than deleted and added  but github does not.

Issue #102, opened 2018-11-22
Result from loops


marpulli:
At the minute the loop does not return anything and there is no quick way to get a solution from a loop. We also can't access the model easily. We should fix this.

PR #110, opened 2018-11-27
get_current_best method implemented for bo

javiergonzalezh:
*Issue #102 , if available:*

*Description of changes:* I added a simple method in the BO loop to have access to the model and the current best found value.

### DIFF
--- a/emukit/bayesian_optimization/loops/bayesian_optimization_loop.py
+++ b/emukit/bayesian_optimization/loops/bayesian_optimization_loop.py
@@ -1,5 +1,9 @@
 # Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
+
+
+import numpy as np
+
 from ..acquisitions import ExpectedImprovement
 from ..acquisitions.log_acquisition import LogAcquisition
 from ..local_penalization_calculator import LocalPenalizationPointCalculator
@@ -28,6 +32,8 @@ class BayesianOptimizationLoop(OuterLoop):
                               Defaults to updating hyper-parameters every iteration.
         """
 
+        self.model = model
+
         if acquisition is None:
             acquisition = ExpectedImprovement(model)
 
@@ -48,3 +54,12 @@ class BayesianOptimizationLoop(OuterLoop):
         loop_state = create_loop_state(model.X, model.Y)
 
         super().__init__(candidate_point_calculator, model_updaters, loop_state)
+
+    def get_current_best(self) -> dict:
+        """
+        Returns a dictionary with the current best found location (minimum) and its value.
+
+        """
+        return {'minimum': self.loop_state.X[np.argmin(self.loop_state.Y),None], 'value':np.min(self.loop_state.Y)}
+
+


emukit/bayesian_optimization/loops/bayesian_optimization_loop.py
```diff
@@ -48,3 +54,12 @@ def __init__(self, model: IModel, space: ParameterSpace, acquisition: Acquisitio
         loop_state = create_loop_state(model.X, model.Y)
 
         super().__init__(candidate_point_calculator, model_updaters, loop_state)
+
+    def get_current_best(self) -> dict:
+        """
+        Returns a dictionary with the current best found location (minimum) and its value.
+
+        """
+        return {'minimum': self.loop_state.X[np.argmin(self.loop_state.Y),None], 'value':np.min(self.loop_state.Y)}
```
marpulli: I'd use `:` rather than `None` as it makes sense of the minimum location to be 1d rather than 2d.

It's not clear "minimum" is the location of the minimum. Maybe "minimum_location" and "minimum_value" are clearer key name?

emukit/bayesian_optimization/loops/bayesian_optimization_loop.py
```diff
@@ -48,3 +54,12 @@ def __init__(self, model: IModel, space: ParameterSpace, acquisition: Acquisitio
         loop_state = create_loop_state(model.X, model.Y)
 
         super().__init__(candidate_point_calculator, model_updaters, loop_state)
+
+    def get_current_best(self) -> dict:
+        """
+        Returns a dictionary with the current best found location (minimum) and its value.
```
marpulli: Could you put the key names in the docstring?

emukit/bayesian_optimization/loops/bayesian_optimization_loop.py
javiergonzalezh: Done

emukit/bayesian_optimization/loops/bayesian_optimization_loop.py
javiergonzalezh: Done

emukit/bayesian_optimization/loops/bayesian_optimization_loop.py
```diff
@@ -55,11 +57,32 @@ def __init__(self, model: IModel, space: ParameterSpace, acquisition: Acquisitio
...
+        self.best_found_value_per_iteration = self._current_best_vector(loop_state.Y)
+
+    def _current_best_vector(self,Y) -> np.array:
+        '''
+        Returns a one-dimensional numpy array whose ith element us the minimum of the components 0:i of Y.
+
+        :param Y: the numpy array with the current best found value per iteration,
+        '''
+
+        Y_best = Y.flatten().copy()
```
marpulli: `return np.minimum.accumulate(Y).flatten()` does this method in one line

emukit/bayesian_optimization/loops/bayesian_optimization_loop.py
```diff
@@ -48,3 +56,33 @@ def __init__(self, model: IModel, space: ParameterSpace, acquisition: Acquisitio
         loop_state = create_loop_state(model.X, model.Y)
 
         super().__init__(candidate_point_calculator, model_updaters, loop_state)
+
+    def get_results(self):
+        return BayesianOptimizationResults(self.loop_state)
+
+
+class BayesianOptimizationResults:
```
marpulli: This class assumes one output dimension. I think this is OK for now but we will need a different class if we do multi-output BO in future.

javiergonzalezh: Sure, for the muti-output dimension case we'll have another loop (based on the notebook you wrote for multi-fidelity) so we can implement an specific method there

emukit/bayesian_optimization/loops/bayesian_optimization_loop.py
javiergonzalezh: thanks! done!

emukit/bayesian_optimization/loops/bayesian_optimization_loop.py
```diff
@@ -45,3 +51,20 @@ def __init__(self, space: ParameterSpace, model: IModel, acquisition: Acquisitio
...
+    def __init__(self,loop_state: LoopState):
```
apaleyes: LoopState class is not imported, and that fails the build

PR #112, opened 2018-11-29
Add model attribute to experimental design loop

marpulli:
*Description of changes:* Add model property to experimental design loop


emukit/experimental_design/model_based/experimental_design_loop.py
```diff
@@ -42,3 +42,5 @@ def __init__(self, space: ParameterSpace, model: IModel, acquisition: Acquisitio
         loop_state = create_loop_state(model.X, model.Y)
 
         super().__init__(candidate_point_calculator, model_updater, loop_state)
+
+        self.model = model
```
mmahsereci: Is there a reason why this is defined after super and the other stuff before?

marpulli: The other stuff is needed to calculate the inputs to super and this does not

Issue #64, opened 2018-11-05
Standard acquisition function test suite


marpulli:
We should be able to have some standard tests for all acquisition functions such as testing output shapes and numerical gradient checks. This would make it easier to create new acquisitions and make sure they are all tested.

PR #118, opened 2018-11-30
Refactor acquisition tests

marpulli:
*Issue #, if available:* #64 

*Description of changes:* Using pytest fixtures and other magic to have standard acquisition tests.

- conftest.py is the place to put fixtures that are shared across test files apparently. I've made a couple of commonly used ones. I can probably use these elsewhere as well if people are happy with this PR.
- To use these tests for a new acquisition you would need to write a new fixture that constructs the acquisition and add it the the list of acquisition function names. That's it.


tests/emukit/test_acquisitions.py
```diff
@@ -0,0 +1,118 @@
...
+from emukit.bayesian_optimization.acquisitions.log_acquisition import LogAcquisition
+from emukit.core import ParameterSpace, ContinuousParameter, InformationSourceParameter
+from emukit.core.acquisition.acquisition_per_cost import CostAcquisition
+
+from emukit.bayesian_optimization.acquisitions import ProbabilityOfImprovement
+from emukit.experimental_design.model_based.acquisitions import ModelVariance, IntegratedVarianceReduction
+from emukit.quadrature.acquisitions import SquaredCorrelation
+
+default_grad_tol = 1e-7
+gradient_acquisitions = [('negative_lower_confidence_bound_acquisition', default_grad_tol),
```
apaleyes: Is there any way these could be named tuples, dictionaries, or anything like that? If we add say one or two more parameters here, it would be next to impossible to guess what each number exactly means. Even 1e-3 does not tell you much (must be a precision? of what?)

tests/emukit/conftest.py
```diff
@@ -0,0 +1,38 @@
+import GPy
```
apaleyes: Can you add a comment on top of this file explaining what is it for, what does it contain?

tests/emukit/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+
+from emukit.core import ContinuousParameter, ParameterSpace
+from emukit.model_wrappers import GPyModelWrapper
+from emukit.model_wrappers.gpy_quadrature_wrappers import convert_gpy_model_to_emukit_model
+from emukit.quadrature.methods import VanillaBayesianQuadrature
+
+
+@pytest.fixture
+def n_dims():
+    # default to 2 dimensional inputs for tests, but can be overridden by individual tests if desired
```
apaleyes: How to override this in a test? Define a fixture with the same name?

tests/emukit/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+@pytest.fixture
+def continuous_space(n_dims):
+    params = [ContinuousParameter('x' + str(i), 0, 1) for i in range(n_dims)]
+    return ParameterSpace(params)
+
+
+@pytest.fixture
+def vanilla_bq_model(gpy_model, continuous_space, n_dims):
```
apaleyes: A bit surprised to see such a specific thing here. Is it common? Where do we use this fixture? If this is for the acquisitions stuff later on, I'd suggest moving it there.

Other fixtures in this file do look very reusable.

tests/emukit/test_acquisitions.py
```diff
@@ -0,0 +1,118 @@
...
+
+@pytest.fixture
+@pytest.mark.parametrize('n_dims', [2])
+def multi_source_entropy_search_acquisition(gpy_model):
+    space = ParameterSpace([ContinuousParameter('x1', 0, 1), InformationSourceParameter(2)])
+    return MultiInformationSourceEntropySearch(gpy_model, space, num_representer_points=10)
+
+
+# Helpers for creating parameterized fixtures
+def create_all_fixtures():
```
apaleyes: A more specific name would be nice, cause this just creates acquisition functions. "all fixtures" is kind of ambiguous

tests/emukit/test_acquisitions.py
marpulli: Done

tests/emukit/conftest.py
marpulli: Moved to the acquisition tests file

tests/emukit/conftest.py
marpulli: Added a comment on one way to do it. There are others, one of which is to override the fixture in the local file with the same name.

tests/emukit/test_acquisitions.py
marpulli: Done

Issue #125, opened 2018-12-03
Mac builds take a long time to complete


apaleyes:
... and sometimes even timeout for transient reasons, succeeding on retry.

Let's try to investigate and see if we can cut this time down. Mark already started in #122

marpulli:
Updating all brew packages seems to take about half of the build time. The python 3.5 build works without updating but 3.6 does not. The build is here:

https://travis-ci.org/amzn/emukit/builds/461907330?utm_source=github_status&utm_medium=notification

marpulli:
The build time has halved with #128, are we happy or do we still want quicker?

apaleyes:
Yes, let's close. osx builds now run just a bit longer than linux, all under 10 minutes, that should be good

PR #128, opened 2018-12-03
Do not merge: investigating mac builds timing further

apaleyes:
*Issue #, if available:* #125 

*Description of changes:* This builds on top of #122 . From [here](https://github.com/pypa/pipenv/issues/2593) is seems as though pyenv on Macs is just behind in terms of python version, so perhaps going down a minor release would help?

ANother option to consider is just "brew update pyenv", which I'll try next.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ python:
   - '3.6'
 before_install: |
  if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
-   brew update
    # Recommended for pyenv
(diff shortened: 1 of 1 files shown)


apaleyes:
Going down to 3.6.5 worked, but does not feel like a right option. So we can still use it, but now another attempt with `brew upgrade pyenv`.

apaleyes:
3.7 build fails while installing GPy. [Known issue](https://github.com/SheffieldML/GPyOpt/issues/193). I guess 3.7 is a topic for another day.

PR #136, opened 2018-12-05
Updated lots of links

apaleyes:
*Issue #, if available:* None

*Description of changes:*  Some minor stuff in readme, most importantly added a link to the website. Also changed all links to relative in the index notebook.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,13 @@
 
 [![Master Branch Build Status](https://travis-ci.org/amzn/emukit.svg?branch=master)](https://travis-ci.org/amzn/emukit) |
 [![Develop Branch Build Status](https://travis-ci.org/amzn/emukit.svg?branch=develop)](https://travis-ci.org/amzn/emukit) |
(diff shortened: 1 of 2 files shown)


mmahsereci:
In the README.md there is still '[coming soon]' for BQ. Could you also delete that in this PR?

I clicked on some links in the index notebook and I can not open them in a new tab which is a bit annoying. Is that related to the relative links?

apaleyes:
Good catch on the note about BQ, removed.

Relative links shouldn't behave any different to absolute ones in this regard. Also, when browsing on GitHub, say [here](https://github.com/amzn/emukit/tree/fbc01990b1b8ff7c69afce0721ad95aa93ac37bb/notebooks), if I open a link it loads in the same tab.

apaleyes:
@mmahsereci spotted an issue: when you "open in a new tab" a link from the index notebook while browsing in github, that gives an error. As a compromise we agreed to give link nbviewer website in README. So this issue will still be there in github, but only if you browse sources. If you, on the other hand, just follow link in README, notebook navigation will work

PR #141, opened 2018-12-07
some fixes to BQ

mmahsereci:
*Issue #, if available:*

*Description of changes:*

- I added a new method to the standard kernel interface. I opted for it since it makes the gradient computation much more efficient going forward. GPy wrappers are changed accordingly. 

- added observation transform to set\_data warped method and Y properties in the warped BQ method. I had missed that before since vanilla BQ has the unit transform. I adjusted the warped method test accordingly since the test mocks the warped method with some data Y and without an explicit implementation of the transform it would throw an error now. I am testing the vanilla BQ instead now.

- I tidied up the predict methods in the base GP and warped GP (not user facing) since they had boolean switches and the input (such as ‘return\_full\_cov=False’) which changed the output. I added separate methods for these cases.

- I tidied up some docstrings and type annotations.

All tests still pass. 


emukit/model_wrappers/gpy_quadrature_wrappers.py
```diff
@@ -143,6 +155,16 @@ def dK_dx1(self, x1: np.ndarray, x2: np.ndarray) -> np.ndarray:
         dK_dx1 = - K[None, ...] * scaled_vector_diff
         return dK_dx1
 
+    def dKdiag_dx(self, x: np.ndarray) -> np.ndarray:
+        """
+        gradient of the diagonal of the kernel (the variance) v(x):=k(x, x) evaluated at x
+
+        :param x: argument of the kernel, shape = (n_points M, input_dim)
```
apaleyes:
```suggestion
        :param x: argument of the kernel, shape (n_points M, input_dim)
```

emukit/quadrature/interfaces/base_gp.py
```diff
@@ -13,6 +13,12 @@ class IBaseGaussianProcess(IModel):
     """Interface for the quadrature base-GP model"""
 
     def __init__(self, kern: QuadratureKernel) -> None:
+        """
+        If this GP is initialized with data, use the raw evaluations Y of the integrand and not transformed values as
+        this is a general class that can be used with various quadrature methods. The transformation will be performed
```
apaleyes: I can't follow this sentence. Can it be simpler?

emukit/quadrature/methods/vanilla_bq.py
```diff
@@ -4,48 +4,67 @@
...
 
-    def inverse_transform(self, Y):
+    def inverse_transform(self, Y: np.ndarray) -> np.ndarray:
         """ Transform from integrand to base-GP """
         return Y
 
-    def predict_base(self, X_pred, return_full_cov=False):
-        """ Computes the predictive mean and covariance """
-        m, cov = self.base_gp.predict(X_pred, return_full_cov)
+    def predict_base(self, X_pred: np.ndarray) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
```
apaleyes: Wander if that should be a private method

mmahsereci: I thought about that, too, then I opted for this because some people might want to e.g. plot the base GP. Otherwise one would need to call `loop.model.base_gp.predict()` instead of `loop.model.predict_base()`.
But I am also fine with changing it, whatever you think is a better user experience.

emukit/model_wrappers/gpy_quadrature_wrappers.py

tests/emukit/quadrature/test_vanilla_bq.py
```diff
@@ -22,3 +25,22 @@ def test_vanilla_bq_transformations():
...
+    Y_train = np.random.rand(5, 1)
+
+    mock_cparam = mock.create_autospec(ContinuousParameter)
+    mock_bounds = mock.create_autospec(IntegralBounds)
+    mock_bounds.convert_to_list_of_continuous_parameters.return_value = 2 * [mock_cparam]
+    mock_kern = mock.create_autospec(QuadratureKernel, integral_bounds=mock_bounds)
+    mock_gp = mock.create_autospec(IBaseGaussianProcess, kern=mock_kern, X=X_train, Y=Y_train)
+    method = VanillaBayesianQuadrature(base_gp=mock_gp)
+
+    assert np.all(method.X == X_train)
```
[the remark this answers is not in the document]
mmahsereci: changed

emukit/quadrature/interfaces/base_gp.py
mmahsereci: I think I can simply just say 'If this GP is initialized with data X, Y, use the raw evaluations Y of the integrand' and skip the explanation? It might cause more confusion that it solves.

mmahsereci:
@apaleyes I am unsure why the checks did not pass. Could you have a look if you find the time?

emukit/model_wrappers/gpy_quadrature_wrappers.py
```diff
@@ -54,15 +54,23 @@ def set_data(self, X: np.ndarray, Y: np.ndarray) -> None:
         """
         self.gpy_model.set_XY(X, Y)
 
-    def predict(self, X_pred: np.ndarray, full_cov=False) -> Tuple[np.ndarray, np.ndarray]:
+    def predict(self, X_pred: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
         """
         Predictive mean and (co)variance at the locations X_pred
```
marpulli:
```suggestion
        Predictive mean and variance at the locations X_pred
```

emukit/model_wrappers/gpy_quadrature_wrappers.py
```diff
@@ -54,15 +54,23 @@ def set_data(self, X: np.ndarray, Y: np.ndarray) -> None:
...
         :param X_pred: points at which to predict, with shape (number of points, dimension)
-        :param full_cov: if True, return the full covariance matrix instead of just the variance
-        :return: Predictive mean, predictive (co)variance
+        :return: Predictive mean, predictive variances shapes (num_points, 1) and (num_points, 1)
+        """
+        return self.gpy_model.predict(X_pred, full_cov=False)
+
+    def predict_with_full_covariance(self, X_pred: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
+        """
+        Predictive mean and (co)variance at the locations X_pred
```
marpulli:
```suggestion
        Predictive mean and covariance at the locations X_pred
```

emukit/model_wrappers/gpy_quadrature_wrappers.py
```diff
@@ -143,6 +151,16 @@ def dK_dx1(self, x1: np.ndarray, x2: np.ndarray) -> np.ndarray:
...
 
+    def dKdiag_dx(self, x: np.ndarray) -> np.ndarray:
+        """
+        gradient of the diagonal of the kernel (the variance) v(x):=k(x, x) evaluated at x
+
+        :param x: argument of the kernel, shape (n_points M, input_dim)
+        :return: the gradient of the diagonal of the kernel evaluated at x, shape (input_dim, M)
+        """
+        num_points, input_dim = x.shape
+        return np.zeros((input_dim, num_points))
```
marpulli: I guess this is correct while we only have stationary covariance functions implemented but in general I think it should call `self.kernel.dKdiag_dx`?

emukit/model_wrappers/gpy_quadrature_wrappers.py

emukit/model_wrappers/gpy_quadrature_wrappers.py

emukit/model_wrappers/gpy_quadrature_wrappers.py
mmahsereci: this is the specific wrapper for the RBF kernel which is a stationary kernel.

marpulli: Ah, so it is, I missed that.

Issue #151, opened 2019-01-07
MultiSourceAcquisitionOptimizer docstring / code mismatch


dekuenstle:
Hi,

in MultiSourceAcquisitionOptimizer.optimize, the best source is determined according to the [argmin of the sources' optimal aquisition values](https://github.com/amzn/emukit/blob/7e9d3ca212e9149a399911c482adf15514168872/emukit/core/optimization/multi_source_acquisition_optimizer.py#L66). The docstrings of both, the function and the class, are "picking the information source where the value of the acquisition at the optimum input location **is highest**." 

Cheers,
David

marpulli:
Hi David,

Thanks for opening this issue! 

I think the confusion comes from the fact that we minimize the negative of the acquisition function and the optimum value returned by `self.acquisition_optimizer.optimize(acquisition, context)` [here](https://github.com/amzn/emukit/blob/7e9d3ca212e9149a399911c482adf15514168872/emukit/core/optimization/multi_source_acquisition_optimizer.py#L64) is actually the negative of the acquisition value at the optimum.

We then take the minimum of the negative acquisition values in order to find the maximum of the non-negative acquisition values. 

I agree that this is confusing and we should change `self.acquisition_optimizer.optimize(acquisition, context)` to return the acquisition value at the optimum rather than the negative acquisition value at the optimum.

Thanks,
Mark

dekuenstle:
Hi Mark,
thanks for the explanation.
As a quickfix, the docstrings of the return value could also be changed to `:return: ... *negative* acquisition value at optimum)` as an alternative to changing the return value itself (which might break people's code).

Cheers
David

PR #152, opened 2019-01-14
Fix comment about acquisition optimizer return values

marpulli:
*Issue #, if available:* #151

*Description of changes:* Correction to comment, we actually return the negative of the acquisition value at the optimum rather than the acquisition value itself.


mmahsereci:
Proposal for preciser language:

- 'maximum/minimum' instead of 'optimum' for the value of the function.
- 'maximizer/minimizer' or 'location of max/min' or 'max/min point' for the location of the optimum. 

In a strict sense 'acquisition  value at optimum' is thus ambiguous in two ways.

Issue #154, opened 2019-01-15
MultiSourceFunctionWrapper result: Wrong x-y-pairs for unsorted sources


dekuenstle:
Hey,
if the input to a MultiSourceFunctionWrapper is not sorted by the source column, 
then the list of results has input in the original order, but output ordered by source.
I did not find any docs or checks about the input order.

```python
X = np.array([[x0, 0], [x1, 1], [x2, 0], [x3, 1]])
f = MultiSourceFunctionWrapper([f0, f1])
results = f.evaluate(X)

# input in results are in the expected order
results[0].X == [x0, 0]
results[1].X == [x1, 1]
...
# output is sorted by source
results[0].Y == f0([x0, 0])
results[1].Y == f0[x2, 0])
results[2].Y == f1([x1, 1])
results[3].Y == f1[x3, 1])
```
In my oppinion, the outputs should be resorted to the input order before [zipping them](https://github.com/amzn/emukit/blob/master/emukit/core/loop/user_function.py#L117). 
I can prepare a PR, if it is welcome.

Cheers, 
David

marpulli:
Good catch, I agree that they should be resorted to the input order. 

A PR would be very welcome, thanks!

PR #156, opened 2019-01-16
Test and fix multi source function wrapper

dekuenstle:
- add tests for multi source function wrapper
- fix reordered outputs described in Issue #154 
- return cost if provided by source functions

### DIFF
--- a/emukit/core/loop/user_function.py
+++ b/emukit/core/loop/user_function.py
@@ -105,15 +105,19 @@ class MultiSourceFunctionWrapper(UserFunction):
 
         _log.info("Evaluating multi-source user function for {} point(s)".format(inputs.shape[0]))
         # Run each source function for all inputs at that source
-        outputs = []
+        indices, outputs = [], []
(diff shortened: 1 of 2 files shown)


dekuenstle:
> Now that we guarantee a certain output order, can you please add this information into the docstring? Otherwise should be good to go

The results' X values (input) were already in this order before, we did not implement an additional order guarantee.  
The sorting of the outputs is just there to enforce matching input-output/cost pairs in the result objects (fix for bug #154 ). UserFunctionWrapper.evaluate also doesn't state a guarantee explicitly even if the results' X values are in the input order, too.

@apaleyes  Would you still explicitly state an guarantee?

apaleyes:
@dekuenstle you are right, guarantee is probably too strong of a word. However now we make sure that the returned results are ordered the same way the input was ordered. Which is just common sense, but I think it may still be a useful note in the docstring.

However it should be fine without it either, so I'll leave it to your own judgement. Approved!

PR #165, opened 2019-02-08
Use robust optimization flag

marpulli:
Small change to non-linear model so that it continues to try to optimize at a new initial point if we get errors during the optimization, usually due to matrices becoming singular.


marpulli:
LinearAlgebra error from numpy, it causes the optimization to stop even if we have specified a number of random restarts. This flag just makes it go onto the next random restart in case of a failure.

Issue #169, opened 2019-02-21
Discrete space optimization


dekuenstle:
There are already discrete and categorical parameters in emukit, but gradient-based acquisition-optimization seems not optimal for such spaces. 
Are there plans for other optimizers like local search?
Probably additional optimizers would require additional features of the parameter space (e.g. value neighbourhood, transform vector encoding to parameter values, inverse transform, …). 

What are your thoughts about this?

Cheers 
David

apaleyes:
As far as I remember, we don't have immediate plans for implementing other optimizers. That may come later, but are not working on anything that requires them at the moment. That said, of course we won't oppose a PR implementing that.

As for parameter space functionality to support specific optimizer, i personally think, unless these parameter manipulations are super general, should be done inside the optimizer itself. But that's just a general thought, specific cases might differ.

marpulli:
I definitely agree that optimizers that treat discrete/categorical parameters properly would be a great addition to emukit. I don't know a lot about the different optimizers we could use or the required additional features you mention. 

Are the transforms required anything like the ones we have implemented in `emukit.core.encodings`? They don't currently have inverse transforms implemented, but that seems like a reasonable addition to these classes.

aaronkl:
AFAIK, the most people use some stochastic local search variants to optimize the acquisition function in discrete spaces (see [here](https://ml.informatik.uni-freiburg.de/papers/11-LION5-SMAC.pdf) for example). I think we could fairly easily implement such as local search method as following:

1. get a starting configuration and the maximum number of steps as input
2. compute the neighborhood of the starting configuration by flipping each element while keeping all other elements of the configuration fix
3. for each neighbor compute the acquisition value
4. if we reached the maximum number of steps or we haven't improved upon our starting point return the starting point  otherwise replace the starting point with the best neighbor and repeat with step 2.

The only somewhat challenging part here is to compute the neighbors. For categorical parameters we could simply encode them first as a one-hot-encoding vector and randomly replace one 0 by 1. For ordinal or integer parameters we would need to change it to the next higher / lower value.
So I guess we could use `emukit.core.encoding` for that, we would just need another function that transforms them back into a Parameter object.

Issue #140, opened 2018-12-07
Dependency on GPy and GPyOpt


apaleyes:
I'd like to open a discussion on Emukit's dependency on GPy and GPyOpt.

GPy dependency causes us some trouble: it is most likely cause of installation failures, it drags in matplotlib and plotly, and it has [known issues](https://github.com/SheffieldML/GPyOpt/issues/193) with Python 3.7. For these reasons I would like to see if we could somehow get rid of that dependency being required.

We depend on GPy and GPyOpt in following places:
1. Model free designs
It is possible to even copy over relevant code pieces from GPyOpt and get rid of that dependency.
2. Optimizer
Same thing, we could copy, or re-impelement
3. Multi-fidelity
This is harder, but we could consider multi-fidelity as an extra bit of the package. People could successfully run decision loops in Emukit without multi-fidelity feature.

Note that 1 and 2 help us remove GPyOpt as a dependency at all.

Opinions welcome.

apaleyes:
Point 1 above is addressed.

marpulli:
We have removed the GPyOpt dependency. Do we actually want to remove the GPy dependency?

apaleyes:
I actually think we do. GPy only comes into play if users want to use multi-fidelity models, so you don't need GPy for core outer loop stuff of the library. Besides we always talk about how we are model agnostic and how emukit can be used with any modelling backend, and yet we install GPy along with it.

Here is the list of dependencies we get installed on a completely new and clean python environment:

```
Installing collected packages: setuptools, numpy, scipy, six, decorator, paramz, kiwisolver, cycler, pyparsing, python-dateutil, matplotlib, retrying, certifi, idna, urllib3, chardet, requests, attrs, pyrsistent, jsonschema, ipython-genutils, traitlets, jupyter-core, nbformat, pytz, plotly, GPy, emcee, emukit
```

Most of that is dragged along by GPy, and I think that's a lot for a relatively tiny library such as emukit.

marpulli:
As well as the multi-fidelity package, we also need some modelling package to run the intro code. Our model for the intro code doesn't necessarily have to be done with GPy, we could use sklearn if we think that's going to be better maintained in the future. We could even just implement a very simple GP model in emukit examples.

apaleyes:
Oh, having a simple GP implementation for basic stuff is a great idea, let's look into that

mmahsereci:
[...]

PR #170, opened 2019-02-25
Discrete optimizer

dekuenstle:
*Issue #169*

*Description of changes:*
As discussed in #169 this PR contains optimizers which handle non-continuous spaces better than the default gradient-based optimizer: 

* Add *local search acquisition optimizer* (using one-exchange neighbours)
* Add *random search acquisition optimizer*
* Rename default *acquisition optimizer* to *gradient acquisition optimizer*
* Add *uniform sampling method* to parameter space
* Add and refactor *acquisition optimizer tests*
* Optional argument for *acquisition optimizer in BO and ED loops*

The local search implementation is currently limited to my needs. These limitations include:

* ~Neighbourhood defined only for ordinal and one-hot categorical parameters, fails for other parameters.  
  If neighbourhood is defined in the parameter classes a more generic implementation than [the current](https://github.com/dekuenstle/emukit/blob/discrete_optimizer/emukit/core/optimization/local_search_acquisition_optimizer.py#L29) would be possible.~
* Full one-exchange neighbourhood is evaluated, might be problematic for e.g. a lot of categories.  
  You can check the implementation of [ConfigSpace](https://github.com/automl/ConfigSpace/blob/master/ConfigSpace/util.py#L79) how non-finite or large neighbourhoods can be handled.
* Search ends in first found (local) minimum. Other convergence criterions could be useful.


emukit/bayesian_optimization/loops/bayesian_optimization_loop.py
```diff
@@ -28,6 +29,7 @@ def __init__(self, space: ParameterSpace, model: IModel, acquisition: Acquisitio
                             size is greater than one, this acquisition must output positive values only.
         :param update_interval: Number of iterations between optimization of model hyper-parameters. Defaults to 1.
         :param batch_size: How many points to evaluate in one iteration of the optimization loop. Defaults to 1.
+        :param acquisition_optimizer: Uses gradient based optimizer if None. Defaults to None.
```
apaleyes: This should really just say "Optimizer blah blah. Defaults to gradient base optimizer." I am not sure if this has any implications for the code, or just a docstring change.

emukit/core/optimization/__init__.py
```diff
@@ -3,4 +3,6 @@
 
 
 from .acquisition_optimizer import AcquisitionOptimizer  # noqa: F401
```
apaleyes: Maybe we should consider renaming AcquisitionOptimizer -> GradientAcquisitionOptimizer ?

emukit/core/optimization/local_search_acquisition_optimizer.py
```diff
@@ -0,0 +1,110 @@
+import logging
+from typing import List, Tuple
+
+import numpy as np
+
+from emukit.core import (CategoricalParameter, OneHotEncoding, OrdinalEncoding,
```
apaleyes: All emukit internal imports should be relative

emukit/core/optimization/local_search_acquisition_optimizer.py
```diff
@@ -0,0 +1,110 @@
...
+from emukit.core.acquisition import Acquisition
+from emukit.core.optimization.acquisition_optimizer import \
+    AcquisitionOptimizerBase
+from emukit.experimental_design.model_free.random_design import RandomDesign
+
+_log = logging.getLogger(__name__)
+
+
+class LocalSearchAcquisitionOptimizer(AcquisitionOptimizerBase):
+    """ Optimizes the acquisition function """
```
apaleyes: This should have much more elaborate description.

emukit/core/optimization/local_search_acquisition_optimizer.py
```diff
@@ -0,0 +1,110 @@
...
+        :param num_steps: Maximum number of steps to follow from each start point.
+        :param num_samples: Number of initial sampled points where the local search starts.
+        """
+        self.space = space
+        self.num_steps = num_steps
+        self.num_samples = num_samples
+        self.sampler = RandomDesign(space)
+
+    def _neighbours_per_parameter(self, all_features: np.ndarray) -> List[np.ndarray]:
+        assert all_features.ndim == 1, "Expected 1d array, got {}d".format(all_features.ndim)
```
apaleyes: For now I am going to assume asserts are temp debugging stuff.

emukit/core/optimization/local_search_acquisition_optimizer.py
```diff
@@ -0,0 +1,110 @@
...
+            if isinstance(parameter, CategoricalParameter):
+                if isinstance(parameter.encoding, OrdinalEncoding):
+                    left_right = np.unique([parameter.encoding.round_row(features - 1),
+                                            parameter.encoding.round_row(features + 1)])
+                    neighbours.append(left_right[left_right != features].reshape(-1, 1))
+                elif isinstance(parameter.encoding, OneHotEncoding):
+                    neighbours.append(parameter.encodings[
+                        (parameter.encodings != features).any(axis=1)])
+                else:
+                    raise NotImplementedError("Parameter encoding {} not supported."
```
apaleyes: NotImplemented has a very specific semantic, see https://stackoverflow.com/questions/44315961/when-to-use-raise-notimplementederror . Consider other built-in exception classes, like ValueError or AttributeError

emukit/core/optimization/local_search_acquisition_optimizer.py
```diff
@@ -0,0 +1,110 @@
...
+                                              .format(type(parameter.encoding)))
+            else:
+                raise NotImplementedError("Parameter type {} not supported."
+                                          .format(type(parameter)))
+            current_feature += parameter.dimension
+        return neighbours
+
+    def _neighbours(self, all_features: np.ndarray) -> np.ndarray:
```
apaleyes: Would be great to have even private methods documented. For example I am immediately clear what this one does, and it does a lot

emukit/core/optimization/local_search_acquisition_optimizer.py
```diff
@@ -0,0 +1,110 @@
...
+        incumbent_value = acquisition.evaluate(x.reshape(1, -1)).item()
+        for step in range(self.num_steps):
+            _log.info("Step {}/{}".format(step + 1, self.num_steps))
+            neighbours = self._neighbours(x)
+            acquisition_values = acquisition.evaluate(neighbours)
+            max_index = np.argmax(acquisition_values)
+            max_neighbour = neighbours[max_index]
+            max_value = acquisition_values[max_index]
+            if max_value < incumbent_value:
+                _log.info("Converged")
```
apaleyes: Converged at ...

emukit/core/optimization/local_search_acquisition_optimizer.py
```diff
@@ -0,0 +1,110 @@
...
+    def optimize(self, acquisition: Acquisition, context: dict = None) -> Tuple[np.ndarray, np.ndarray]:
+        """
+        Optimizes the acquisition function, taking into account gradients if it supports them
+        :param acquisition: The acquisition function to be optimized
+        :param context: Optimization context.
+                        Determines whether any variable values should be fixed during the optimization
+        :return: Tuple of (location of maximum, acquisition value at maximizer)
+        """
+        if context is not None:
+            raise NotImplementedError("Handling context is currently not supported.")
```
apaleyes: I don't think this warrants an error. A log warning max

emukit/core/optimization/random_search_acquisition_optimizer.py
```diff
@@ -0,0 +1,42 @@
...
+from emukit.core.acquisition import Acquisition
+from emukit.core.optimization.acquisition_optimizer import \
+    AcquisitionOptimizerBase
+from emukit.experimental_design.model_free.random_design import RandomDesign
+
+_log = logging.getLogger(__name__)
+
+
+class RandomSearchAcquisitionOptimizer(AcquisitionOptimizerBase):
+    """ Optimizes the acquisition function """
```
apaleyes: Ditto

tests/emukit/core/test_optimization.py
```diff
@@ -40,3 +47,49 @@ def test_multi_source_acquisition_optimizer():
     opt_x, opt_val = optimizer.optimize(acquisition)
     np.testing.assert_array_equal(opt_x, np.array([[0., 1.]]))
     np.testing.assert_array_equal(opt_val, np.array([[2.]]))
+
+
+def test_random_search_acquisition_optimizer():
```
apaleyes: I think this file is poorly named. Feel free to rename and create separate ones for different optimizers

emukit/core/optimization/local_search_acquisition_optimizer.py
marpulli: I think I disagree. I'd be kind of annoyed if I passed a context in, waited a while for the loop to converge then found my context was completely ignored.

emukit/core/optimization/local_search_acquisition_optimizer.py
```diff
@@ -0,0 +1,110 @@
...
+
+        parameters = self.space.parameters
+        neighbours = []
+        current_feature = 0
+        for parameter in parameters:
+            features = parameter.round(
+                all_features[current_feature:(current_feature + parameter.dimension)]
+                .reshape(1, -1)).ravel()
+            if isinstance(parameter, CategoricalParameter):
+                if isinstance(parameter.encoding, OrdinalEncoding):
```
marpulli: Shall we move this to a `get_neighbours` method of the encoding class? Also, should we be able to deal with discrete parameters in this bit too?

emukit/core/optimization/local_search_acquisition_optimizer.py
apaleyes: Perhaps, i can see a point in this too. Not sure what to do best, let me think more and get back to this point later

emukit/core/optimization/local_search_acquisition_optimizer.py
apaleyes: Not too sure here. This method seems very specific, having many specific methods like that in Encoding or other general purpose interfaces may make it hard to provide custom implementations of these classes later

emukit/core/optimization/local_search_acquisition_optimizer.py
dekuenstle: Warning logs might get lost but your results change if context is (not) handled. Also it's more pythonic to be explicit here imo.  
If a user does not care about that much about the context, a fallback can be used.
```python
try:
    optimizer.optimize(acquisition, context)
except NotImplementedError:
     _log.warning("...")
    optimizer.optimize(acquisition, None)
```

emukit/bayesian_optimization/loops/bayesian_optimization_loop.py
dekuenstle: I extended all the docstrings.

emukit/core/optimization/local_search_acquisition_optimizer.py
dekuenstle: Sure, i changed them all to relative imports.

emukit/core/optimization/local_search_acquisition_optimizer.py
dekuenstle: They where left from development. I removed them.

emukit/core/optimization/local_search_acquisition_optimizer.py
dekuenstle: I made these logs log more information, but made them debug logs. 
Now the local search optimizer behaves like the default one and logs to info only when it starts optimizing.

emukit/core/optimization/local_search_acquisition_optimizer.py
dekuenstle: Imo core modules should only depend on other core modules. Currently both, random and local search, use the RandomDesign of the experimental_design module. Shall we create a optimizer module outside of core or move the random sampling to core, too?

emukit/core/optimization/local_search_acquisition_optimizer.py
dekuenstle: I've now use TypeError here.

emukit/core/optimization/local_search_acquisition_optimizer.py
dekuenstle: Initially I prefered moving to the parameter. But I cannot think of many more usecases for one-exchange neighbourhoods. Therefore I would prefer keeping this in the optimizer for now and move them to a separate util class/module if more usecases occur.

emukit/core/optimization/__init__.py
dekuenstle: This is a better choice. Are you fine with such breaking changes?

emukit/core/optimization/local_search_acquisition_optimizer.py
dekuenstle: > Also, should we be able to deal with discrete parameters in this bit too?
Discrete ones could be handled like ordinal (select the one lower and higher).

We also should be able to handle continuous parameters: E.g sample k points from a normal distribution (mean at the current, std as hyperparameter e.g. 0.2).

emukit/core/optimization/__init__.py
apaleyes: Yes, I think it should be fine

emukit/core/optimization/local_search_acquisition_optimizer.py
apaleyes: `emukit.optimization`? At the first glance that does not sound bad. @marpulli what do you think?

One tiny caveat is that we still need optimization interfaces in core. So one way would be to keep base class in core, and move the actual implementations to emukit.optimization

marpulli: The original idea was that non-core modules should rely on core only and not any other non-core module, right? `emukit.optimization` would rely on `emukit.experimental_design` which violates that idea. Do we still care? If we do then I guess the answer is to move random design to core.

emukit/core/optimization/local_search_acquisition_optimizer.py
dekuenstle: Discrete and continuous parameters are now also handled.

tests/emukit/core/test_optimization.py
dekuenstle: They are now all separated.

emukit/core/optimization/local_search_acquisition_optimizer.py
dekuenstle: I kept all the optimizers in `core.optimization`. 

Several modules use the RandomDesign to generate random design. Therefore I decided to add a `sample_uniform` method directly to the parameter space to decouple them.

emukit/core/optimization/__init__.py
dekuenstle: It's renamed like suggested.

dekuenstle:
I'm now happy with the features of this PR, so I would remove the Draft-status. 
Or did I miss any?

apaleyes:
@dekuenstle yes, feel free to remove the Draft status. This PR looks like a ton of work, much appreciated!

I still need to do an in-depth review of this

emukit/core/optimization/random_search_acquisition_optimizer.py
```diff
@@ -0,0 +1,48 @@
...
+from .. import ParameterSpace
+from ..acquisition import Acquisition
+from ..optimization.acquisition_optimizer import AcquisitionOptimizerBase
+
+_log = logging.getLogger(__name__)
+
+
+class RandomSearchAcquisitionOptimizer(AcquisitionOptimizerBase):
+    """ Optimizes the acquisition function by evaluating at random points."""
+    def __init__(self, space: ParameterSpace, num_samples: int, **kwargs) -> None:
```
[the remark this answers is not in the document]
dekuenstle: Fixed this.

emukit/core/optimization/local_search_acquisition_optimizer.py
```diff
@@ -0,0 +1,174 @@
...
+                    neighbours.append(parameter.encodings[
+                        (parameter.encodings != features).any(axis=1)])
+                else:
+                    raise TypeError("{} not a supported parameter encoding."
+                                    .format(type(parameter.encoding)))
+            elif isinstance(parameter, DiscreteParameter):
+                current_index = np.argmin(np.abs(
+                    np.subtract(parameter.domain, np.asscalar(features))))
+                this_neighbours = []
+                if current_index > 0:
```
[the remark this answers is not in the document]
dekuenstle: Good catch, thank you! I fixed this and extended the test case.

dekuenstle:
> Thanks for this PR! Sorry this has taken so long for me to review.
> 
> I have a couple of small comment suggestions and a question about the local search for discrete parameters, but apart from that I'm happy.

Thanks for the review! I added the suggested changes.

emukit/core/optimization/local_search_acquisition_optimizer.py
```diff
@@ -0,0 +1,176 @@
+import logging
+from typing import Sequence
```
apaleyes: This can be a single line

    from typing import Sequence, List,...

emukit/core/optimization/local_search_acquisition_optimizer.py
```diff
@@ -0,0 +1,176 @@
...
+from ..optimization.acquisition_optimizer import AcquisitionOptimizerBase
+
+_log = logging.getLogger(__name__)
+
+
+class LocalSearchAcquisitionOptimizer(AcquisitionOptimizerBase):
+    """ Optimizes the acquisition function by greedily following one-exchange neighbours
+        of random initial points.
+
+    One-exchange neighbourhood is defined per parameter type:
```
apaleyes: Seems as though other encodings are not supported. Can you mention that in the docstring, as well as a brief comment on how non-categoricals are handled?

Also, is there a paper or another well accepted source which we could reference?

emukit/core/optimization/local_search_acquisition_optimizer.py
```diff
@@ -0,0 +1,176 @@
...
+      :Categorical parameter with one-hot encoding: All other categories
+      :Categorical parameter with ordinal encoding: Only preceeding and following categories
+    """
+    def __init__(self, space: ParameterSpace, num_steps: int, num_samples: int,
+                 std_dev: float = 0.02, num_continuous: int = 1) -> None:
```
apaleyes: Where did these defaults come from?

emukit/core/optimization/local_search_acquisition_optimizer.py
```diff
@@ -0,0 +1,176 @@
...
+            else:
+                incumbent_value = max_value
+                x = max_neighbour
+        _log.debug("End at step limit with acquisition={:.4f} at {}"
+                   .format(incumbent_value, str(x)))
+        return x, incumbent_value
+
+    def optimize(self, acquisition: Acquisition, context: dict = None) -> Tuple[np.ndarray, np.ndarray]:
+        """
+        Optimizes the acquisition function, taking into account gradients if it supports them
```
apaleyes: I don't think this one actually takes gradients into account

emukit/core/optimization/random_search_acquisition_optimizer.py
```diff
@@ -0,0 +1,52 @@
...
+
+_log = logging.getLogger(__name__)
+
+
+class RandomSearchAcquisitionOptimizer(AcquisitionOptimizerBase):
+    """ Optimizes the acquisition function by evaluating at random points."""
+    def __init__(self, space: ParameterSpace, num_samples: int) -> None:
+        """
+        :param space: The parameter space spanning the search problem.
+        :param num_samples: Number of random sampled points which are evaluated per optimization.
```
apaleyes: I find "num_samples" name confusing in this context. Maybe "num_eval_points"? Very subjective though

emukit/core/optimization/random_search_acquisition_optimizer.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def optimize(self, acquisition: Acquisition, context: dict = None) -> Tuple[np.ndarray, np.ndarray]:
+        """
+        Optimizes the acquisition function, taking into account gradients if it supports them
+        :param acquisition: The acquisition function to be optimized
+        :param context: Optimization context.
+                        Determines whether any variable values should be fixed during the optimization
+        :return: Tuple of (location of maximum, acquisition value at maximizer)
+        """
+        if context is not None:
```
apaleyes: This bit with context is repeated across all optimizers. Perhaps move it to base class, thus making the base class an actual class and making my previous comment invalid?

emukit/core/parameter_space.py
```diff
@@ -147,3 +147,17 @@ def check_points_in_domain(self, x: np.ndarray) -> np.ndarray:
             in_domain = np.all([in_domain, param_in_domain], axis=0)
             encoding_index += param.dimension
         return in_domain
+
+    def sample_uniform(self, point_count: int) -> np.ndarray:
```
apaleyes: This is a brilliant way to resolve our problems with dependency on Random design everywhere. Thanks!

tests/emukit/core/optimization/test_local_search_acquisition_optimizer.py
```diff
@@ -0,0 +1,68 @@
+import numpy as np
```
apaleyes: Few tests for exceptions would be nice

emukit/core/optimization/gradient_acquisition_optimizer.py
```diff
@@ -0,0 +1,81 @@
...
+from ..acquisition import Acquisition
+from .acquisition_optimizer import AcquisitionOptimizerBase
+
+
+import logging
+_log = logging.getLogger(__name__)
+
+
+class GradientAcquisitionOptimizer(AcquisitionOptimizerBase):
+    """ Optimizes the acquisition function """
```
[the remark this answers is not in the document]
dekuenstle: I just realized, that you could also pass the `optimizer` argument and use DIRECT or CMA with GPyOpt instead of L-BFGS. Both  DIRECT/CMA are gradient-free. Therefore this optimizer might be poorly named. What these three have in common, is that they expect a continuous search space. I see three options, which one do you prefer?
1. Rename to `ContinuousAcquisitionOptimizer`
2. Rename to `GPyOptAcquisitionOptimizer`
3. Force use of L-BFGS.

emukit/core/parameter_space.py
dekuenstle: Thank you!

emukit/core/optimization/gradient_acquisition_optimizer.py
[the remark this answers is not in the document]
apaleyes: Direct and CMA are not stable in GPyOpt, let's not even give the users a way to use them. So option 3, thanks for checking

emukit/core/optimization/local_search_acquisition_optimizer.py
dekuenstle: True, this is a c/p mistake.

emukit/core/optimization/random_search_acquisition_optimizer.py
dekuenstle: True these are poorly named. same for the local search.

emukit/core/optimization/local_search_acquisition_optimizer.py
dekuenstle: I extended the docs describing where it comes from and how parameters are handled. Please check, if it's understandable.

emukit/core/optimization/local_search_acquisition_optimizer.py
dekuenstle: Described in the class docstring now.

emukit/core/optimization/random_search_acquisition_optimizer.py
dekuenstle: Please check my latest commit. I moved most of the context handling to the base class. This also removes the direct dependency on GPyOpt from most optimizer files (regarding #140 ).

emukit/core/optimization/acquisition_optimizer.py
```diff
@@ -26,72 +24,3 @@ def optimize(self, acquisition: Acquisition, context: dict = None) -> Tuple[np.n
         :return: Tuple of (location of maximum, acquisition value at maximizer)
         """
         pass
-
-
-class AcquisitionOptimizer(AcquisitionOptimizerBase):
```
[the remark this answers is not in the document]
dekuenstle: The base class can provide common functionality. See my latest commit and my comment below.

emukit/core/optimization/random_search_acquisition_optimizer.py
apaleyes: Neat, thanks! It broke the build though, but should be a simple enough fix.

tests/emukit/core/optimization/test_local_search_acquisition_optimizer.py
dekuenstle: I've added some.

emukit/core/optimization/random_search_acquisition_optimizer.py
dekuenstle: Sorry, I forgot adding the context manager file. It simply wraps the GPyOpt one and should make it straight forward replacing by an implementation without GPyOpt in future PRs.

PR #171, opened 2019-02-27
some fixes to BQ

mmahsereci:
*Issue #, if available:*

*Description of changes:*

1. a bug fix in circular_gaussian test function
2. integrate method of vanilla BQ returns tuple of two floats now instead of tuple of float and 1x1 array (I thought it did that before already...)
3. added shape and some type tests for quadrature kernel and integration model because of 2.

### DIFF
--- a/emukit/quadrature/methods/vanilla_bq.py
+++ b/emukit/quadrature/methods/vanilla_bq.py
@@ -60,7 +60,7 @@ class VanillaBayesianQuadrature(WarpedBayesianQuadratureModel):
         """
         integral_mean, kernel_mean_X = self._compute_integral_mean_and_kernel_mean()
         integral_var = self.base_gp.kern.qKq() - np.square(lapack.dtrtrs(self.base_gp.gram_chol(), kernel_mean_X.T,
-                                                           lower=1)[0]).sum(axis=0, keepdims=True).T
+                                                           lower=1)[0]).sum(axis=0, keepdims=True)[0][0]
         return integral_mean, integral_var
 
     # helpers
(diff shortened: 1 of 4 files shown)


emukit/quadrature/methods/vanilla_bq.py
```diff
@@ -60,7 +60,7 @@ def integrate(self) -> Tuple[float, float]:
         """
         integral_mean, kernel_mean_X = self._compute_integral_mean_and_kernel_mean()
         integral_var = self.base_gp.kern.qKq() - np.square(lapack.dtrtrs(self.base_gp.gram_chol(), kernel_mean_X.T,
-                                                           lower=1)[0]).sum(axis=0, keepdims=True).T
```
apaleyes: If that was indeed a 1x1 array, this `T` is just too charming

emukit/test_functions/quadrature_functions.py
```diff
@@ -31,15 +31,18 @@ def _hennig1D(x: np.ndarray) -> np.ndarray:
 
 
 def circular_gaussian(mean: np.float = 0., variance: np.float = 1.) -> Tuple[UserFunctionWrapper, Callable]:
-    """ 2D toy integrand that is a Gaussian on a circle
     """
-    return UserFunctionWrapper(_circular_gaussian), lambda x: _circular_gaussian(x, mean, variance)
+    2D toy integrand that is a Gaussian on a circle
+    """
+    func = lambda x: _circular_gaussian(x, mean, variance)
+    return UserFunctionWrapper(func), func
```
apaleyes: `UserFunctionWrapper` has `.f` field, so returning `func` is probably redundant

mmahsereci: I did this as a kind of curtesy to users. For instance in case they want to plot the test function, they do not need to figure out how to call it though the wrapped version (might be a bit confusing). 

Our test functions at the moment seem to have various return structures. Some return the wrapped function, some the unwrapped function, and some return the parameter space in addition. Do you think it would be beneficial to have a common structure for all of them?

apaleyes: Agreed. Let's open an issue about it to host the discussion

mmahsereci: OK, so I'll leave it like that for now since changing might involve changing notebooks etc.

Issue #172, opened 2019-02-28
Unify return structure of test function


mmahsereci:
Our test functions at the moment have various return structures. Some return the wrapped function, some the unwrapped function, and some return the parameter space in addition. Should we unify this?

mmahsereci:
I will start doing this sometime soon. Any preferences? Of not I will go for 

return func_wrapped, space, func

for quadrature:

return func_wrapped, integral_bounds, func

apaleyes:
Personally, I don't see much value in returning both func_wrapped and func, as func is available as a property of func_wrapped. But if others agree this is a useful convenience, i won't disagree for long.

mmahsereci:
This is kind of done. Any more comments here?

apaleyes:
Went through our test functions, and they all seem coherent in what they return: either function and space or function and integ bounds. Odd one out is sensitivity, but that's a different story. Thus agreed, let's close this!

PR #173, opened 2019-03-11
cleaning up bq test functions

mmahsereci:
*Issue #172 , if available:* 

*Description of changes:*
For bq functions only. Will do the others in another PR. Return structure is: wrapped_func, integral_bounds
Also, I added one more bq test function.


mmahsereci:
Good point. Nevertheless, integral_bounds is just a list of Tuples, e.g., [(-3, 3), (-4, 2), ...(-7.5, 1.2)].
So one can either input this into the function or just override the output. There is no magic happening with respect to the integral bounds inside the test functions, i.e., the integrals bounds the user puts in would be the integral bounds which are returned by test function. What is the better option?

apaleyes:
@mmahsereci how an "override output" code would look like?

mmahsereci:
@apaleyes , I added the docstrings as discussed.

emukit/test_functions/quadrature/baselines.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+from typing import Tuple, List
+import numpy as np
+from scipy.integrate import quad, dblquad
+
+
+def univariate_approximate_ground_truth_integral(func, integral_bounds: Tuple[float, float]):
+    """
+    Uses scipy.integrate.quad to estimate the ground truth integral
```
apaleyes: Minor: this gives out implementation detail. Better phrasing would be "Estimate the ground truth integral." But for test functions this is very minor concern

emukit/test_functions/quadrature/baselines.py
```diff
@@ -0,0 +1,40 @@
...
+    :returns: integral estimate, output of scipy.integrate.quad
+    """
+    lower_bound = integral_bounds[0]
+    upper_bound = integral_bounds[1]
+    return quad(func, lower_bound, upper_bound)
+
+
+def bivariate_approximate_ground_truth_integral(func, integral_bounds: List[Tuple[float, float]]):
+    """
+    Uses scipy.integrate.dblquad to estimate the ground truth integral
```
apaleyes: ditto

emukit/test_functions/quadrature/baselines.py
mmahsereci: sure, I'll rephrase.

PR #175, opened 2019-03-15
mutual information acquisition for BQ

mmahsereci:
*Issue #, if available:*

*Description of changes:*
The mutual information acquisition for BQ. 


tests/emukit/quadrature/test_quadrature_acquisitions.py
```diff
@@ -0,0 +1,89 @@
...
+ABS_TOL = 1e-4
+
+
+@pytest.fixture
+def model():
+    rng = np.random.RandomState(42)
+    x_init = rng.rand(5, 2)
+    y_init = rng.rand(5, 1)
+
+    gpy_kernel = GPy.kern.RBF(input_dim=x_init.shape[1])
```
javiergonzalezh: In the tests, an alternative to building a model is to use Mock. It is Ok in this case as the model is pretty though.

emukit/quadrature/acquisitions/mutual_information.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class MutualInformation(Acquisition):
+    """
+    This acquisition function is the mutual information between the integral and the new point(s) under a GP-model.
+
+    MutualInformation is a monotonic transform of SquaredCorrelation and hence yields the same acquisition policy in
+    the case of vanilla-BQ.
+
+    .. math::
```
javiergonzalezh: Maybe add a citation here?

emukit/quadrature/acquisitions/mutual_information.py
```diff
@@ -0,0 +1,60 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+
+import numpy as np
+from typing import Tuple
+
+from emukit.core.acquisition import Acquisition
```
marpulli: I think we should use relative imports for imports from emukit, but maybe @apaleyes can clarify?

mmahsereci: I think you are right. I will change that.

emukit/quadrature/acquisitions/mutual_information.py
mmahsereci: Yes, I'll add when it is available :)

Issue #130, opened 2018-12-04
Integration against Gaussian probability measure for BQ


mmahsereci:
so far BQ only integrates against the uniform measure. Integration against a Gaussian would be a nice addition.

mmahsereci:
This was done in the linked PR. I am closing the issue.

PR #176, opened 2019-03-18
Uniform integration measure for BQ

mmahsereci:
*Issue #130 , if available:*

*Description of changes:*
Structure for vanilla BQ integration against some measure. This PR only contains the uniform measure for simplicity. Gaussian will come later.


mmahsereci:
Some additional explanation (to avoid confusion): The PR is also supposed to ensure that whenever the integration bounds are changed, it does so consistently, which has not been the case before. This is why there are some changes in other objects that use the integral bounds. There are also some tests which check if integral bounds are re-set correctly. All this was needed because the integration over the uniform measure is very easy to do by changing the integral bounds temporarily.

emukit/quadrature/methods/vanilla_bq.py
```diff
@@ -52,18 +53,55 @@ def predict_base_with_full_covariance(self, X_pred: np.ndarray) -> Tuple[np.ndar
...
+        # This is equivalent (up to the uniform density) to integrating with respect to the uniform measure.
+        integral_bounds = self.integral_bounds.bounds
+        uniform_bounds = measure.bounds
+        old_integral_bound_list = self.integral_bounds.bounds.copy()
+        new_integral_bound_list = [(max(int_bounds[0], uni_bounds[0]), min(int_bounds[1], uni_bounds[1]))
+                                   for int_bounds, uni_bounds in zip(integral_bounds, uniform_bounds)]
+
+        # setting new bounds also checks bound validity
+        self.integral_bounds = new_integral_bound_list
+        integral_mean_lebesgue, integral_var_lebesgue = self._integrate_lebesgue()
```
marpulli: The other approach would be to input `integration_bounds` to `self._integrate_lebesgue()` and then into all kernel methods that require the bounds instead of having the bounds as an attribute on the kernel. This would avoid needing to worry about the integration bounds setting and resetting. Thoughts?

mmahsereci: I see your point, but I think that might cause bigger problems because then you need to make sure that whoever designs any integration method use the same bounds for all the methods of the kernel, which is very error prone. Having the attribute in the kernel (the only place where integration happens) avoids that. I'd rather make sure to write a nice setter.

Also, to be clear, I do not want the user to change the integral bounds. They are the parameter space and I see no point in changing them. It is just a way to internally do the uniform measure.

PR #177, opened 2019-03-20
Add n_restarts option to gpy model wrapper

marpulli:
Allows user to do multiple restarts for optimization of hyper-parameters for gpy models

### DIFF
--- a/emukit/model_wrappers/gpy_model_wrappers.py
+++ b/emukit/model_wrappers/gpy_model_wrappers.py
@@ -16,8 +16,13 @@ class GPyModelWrapper(IModel, IDifferentiable, ICalculateVarianceReduction, IEnt
     """
     This is a thin wrapper around GPy models to allow users to plug GPy models into Emukit
     """
-    def __init__(self, gpy_model):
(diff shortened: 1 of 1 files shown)


emukit/model_wrappers/gpy_model_wrappers.py
```diff
@@ -48,7 +53,7 @@ def optimize(self):
         """
         Optimizes model hyper-parameters
         """
-        self.model.optimize()
+        self.model.optimize_restarts(self.n_restarts, robust=True)
```
mmahsereci: Is optimize() and optimize_restarts(1) the same? It might differ in case one is deterministic and the other one random.

marpulli: Yep, it's the same. 

[Here](https://github.com/sods/paramz/blob/master/paramz/model.py#L170) is the code that does the randomization, it only randomizes before the second and subsequent restarts.

mmahsereci: 👍

PR #178, opened 2019-03-20
Add multi-fidelity test functions

marpulli:
Adds some multi-fidelity functions used in the multi-fidelity deep GP paper. This does remove the borehole function as an example in it's own right - I think this OK as it's a bit of a boring function and I'm not aware of it's use outside of multi-fidelity.


marpulli:
Added type annotations for output values.

PR #180, opened 2019-03-26
batch prediction ; return var instead of std

aaronkl:
This PR fixes two problems of the random forest (RF):
 - At the moment the RF contains a python for loop to compute the prediction for each test point. However, sklearn allows to make batch prediction which is much more efficient.
 - Besides that, right now the RF returns the standard deviation instead of the variance.


apaleyes:
Travis built completed successfully, but the PR status did not change. Hopefully that was some Github local hiccup. I've added a minor commit to trigger the build again

PR #186, opened 2019-04-11
Fix empty neighbourhood of local parameter, test this

dekuenstle:
*Description of changes:*

Handle the edge case of empty neighbourhood of a discrete parameter.
The other parameters already handled it correctly. 
Occurs rarely in practice, but should not throw an error. 


dekuenstle:
Please don't merge, I realized it's still wrong

dekuenstle:
Thanks for waiting. 
The first fix was introducing an bug if the empty neighbourhood is not of the last parameter.
The discrete parameter was the only one, which made problems with discrete neighbourhoods, the other ones where already fine.

PR #191, opened 2019-04-29
Feature: Information Source Encoding Types

dekuenstle:
Currently information sources are always represented as special discrete parameters.
This is might be appropriate e.g. for most multi-fidelity use cases.
When information sources are not ordered (e.g. multiple dataset) it might not be desired to encode them as (0, 1, 2, ...). This encoding is used to train the model, but some models might infer dataset 0 is more similar to dataset 1 than to dataset 2. 
Therefore this PR proposes to separate encoding and source indices of information source parameters.

*Description of changes:*

* Add ordinal, normal and discrete information source parameters
* Separate source encoding and indices in multisource userfunction and multi source optimizer 
* Fix userfunction, notebook and tests appropriately

### DIFF
--- a/emukit/bayesian_optimization/acquisitions/entropy_search.py
+++ b/emukit/bayesian_optimization/acquisitions/entropy_search.py
@@ -7,7 +7,7 @@ from typing import Union, Callable
 import scipy
 import numpy as np
 
-from ...core import InformationSourceParameter
+from ...core import InformationSourceParameter, DiscreteInformationSourceParameter, OrdinalInformationSourceParameter
 from ...core.acquisition import Acquisition
 from ...core.interfaces import IModel
 from ...core.parameter_space import ParameterSpace
@@ -240,8 +240,13 @@ class MultiInformationSourceEntropySearch(EntropySearch):
 
         # Assume we are in a multi-fidelity setting and the highest index is the highest fidelity
         if target_information_source_index is None:
-            target_information_source_index = max(info_source_parameter.domain)
-        self.target_information_source_index = target_information_source_index
+            if isinstance(info_source_parameter, (DiscreteInformationSourceParameter,
+                                                  OrdinalInformationSourceParameter)):
+                self.target_information_source_index = info_source_parameter.bounds[0][1]
+            else:
+                raise ValueError('target_information_index is None and cannot automatically determined.')
+        else:
+            self.target_information_source_index = target_information_source_index
 
         # Sampler of representer points should sample x location at the target information source only so make a
         # parameter space without the information source parameter
--- a/emukit/core/__init__.py
+++ b/emukit/core/__init__.py
@@ -3,7 +3,8 @@
 
 from .categorical_parameter import CategoricalParameter  # noqa: F401
 from .continuous_parameter import ContinuousParameter  # noqa: F401
-from .discrete_parameter import DiscreteParameter, InformationSourceParameter  # noqa: F401
+from .discrete_parameter import DiscreteParameter  # noqa: F401
 from .encodings import OneHotEncoding, OrdinalEncoding  # noqa: F401
+from .information_source_parameter import *  # noqa: F401, F403
 from .parameter import Parameter  # noqa: F401
 from .parameter_space import ParameterSpace  # noqa: F401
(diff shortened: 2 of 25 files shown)


dekuenstle:
> There might be small comments here and there, but we'll bother about that later. For now I'd like to discuss the whole premise here. 

Yes, I just posted the version I'm currently working with as a suggestion.
Other designs might fit better to the library. 

> Do we actually need all these parameter types? That starts to feel very specific, to the point when I am not sure anymore if this is really necessary.

Definitly, I'm also not happy with all these specific classes.
First I thought about a information source wrapper, which mostly leads through the wrapped parameters attributes (via __getattr__, __setattr__). I cancelled this approach mainly because of two reasons:
1. I was not sure how this affects the type testings (especially isinstance). 2. You would have to transform back from the model encoded values to the parameter values, which is currently not available in a generalized api and requires individual code  (see source_indices methods in the code).

> 
> The only reason InfoSourceParam exists (to my understanding) is so that we could distinguish it among other parameters in multi-fidelity setting. Perhaps a different approach could be used in the library, when we distinguish these params by name (just for example), and then just use discrete or cateogorical params as info source params as necessary? On one hand, it seems less complex, on the other hand, we might get into dangerous string matching before long with this.

In general I *strongly* prefer typed solutions against matching some magic-number or magic-string.
A pythonic solution would use duck-typing to determine a objects purpose, I think.
That said, another solution would be to not having any source-type or -name at all, but let the user
 specify one or multiple parameters by instance or name when calling a multi fidelity/source function.

> @dekuenstle @marpulli @mmahsereci @javiergonzalezh please offer opinions!

marpulli:
Maybe we can just input the encoding into the information source parameter object:

```python
class InformationSourceParam(CategoricalParameter):
    def __init__(self, encoding: Encoding):
        if encoding is None:
            # default the encoding to ordinal - I don't know how it knows the number of information sources here though
        CategoricalParameter.__init__(self, 'source', encoding)

    def source_indices(self, values: np.ndarray) -> np.ndarray:
        """ See InformationSourceParameter.source_indices """
        return np.asarray([self.encoding.get_category(row) for row in self.round(values)])

    @property
    def n_sources(self) -> int:
        """ Number of sources. """
        return self.encodings.shape[0]
```

Then in something like `MultiInformationSourceEntropySearch` where you check what type of `InformationSourceParameter` you have, you could check the type of the encoding object or just assume the target is always to last information source.

This doesn't deal with the `DiscreteInformationSourceParameter` but I'm not 100% convinced that class is necessary and if it is, could we use a different Encoding class to do the same thing? 

Does this make sense?

dekuenstle:
@marpulli In think almost every parameter type could be used to represent multi-information sources. You could also think about e.g. relative dataset size as a continuous fidelity measurement. I'm not saying emukit has to support all the cases right now, but it would be nice if it's easy to extend.

marpulli:
Sorry for the delay in replying. 

I think that continuous vs discrete information sources need different enough treatment that they can be separate classes. For instance, the optimization procedure is likely to be different and evaluating the objective function can be represented as a list of function for discrete information sources but not for continuous. 

All the discrete/categorical information source parameters just refer to an index into a set of information sources. With the current implementation, if the model needs this index to be encoded in a particular way, it can be done in the model itself or in the model wrapper. I see the convenience of not having to encode this by hand in the model but it would be nice to have one class that represents all these so that if we add different types of encodings of categorical variables, we don't need to add a new information source parameter class.

dekuenstle:
Thanks for your answer @marpulli , it helped me sorting thoughts about the topic:

I have cases where not only the model but also the optimizer requires specific encoding.
A viable solution for my case might be to replace the information source in the parameter space with another parameter and write a wrapper for the user function which transforms the other parameter to the information source. Others might only require a wrapper for the model, like you've proposed.

In practice I've realized, that the approach of this PR currently doesn't work completely. E.g. the context handling (using GPyOpt Context Handler) cannot be used for one-hot-parameters without a bunch of specific glue code.

As a conclusion:
Continuous parameters might already be usable with the continuous parameter, if special handling is required it can be done through the parameter index. Anyway continuous and discrete information sources require separater handling. Discrete information sources can be improved by exchangeable encoding. However this requires features which have to be implemented before (e.g. DiscreteParameter uses a Encoding object like CategoricalParameter, Context handling works for all encodings). Therefore I'll close this PR for now.

PR #192, opened 2019-05-03
fix notebook for emukit with categorical variables

aaronkl:
The current notebook crashes since it tries to use a gradient based optimizer per default, which due to the random forest and the categorical space, does not work.
I changed the optimizer to the LocalSearchAcquisitionOptimizer which is applicable here


aaronkl:
yes I found a way to suppress all TF warnings

PR #195, opened 2019-05-14
Cost bo loop

aaronkl:
Currently, we can only pass initial data points and their corresponding function values to the BO optimization loop. However, in case we also want to keep track of the cost per function evaluation, we should also be able to pass the initial cost. Otherwise, this will mess up our internal loop state.


emukit/core/loop/loop_state.py
```diff
@@ -59,6 +59,7 @@ def create_loop_state(x_init: np.ndarray, y_init: np.ndarray, cost: np.ndarray =
 
     :param x_init: x values for initial function evaluations.
     :param y_init: y values for initial function evaluations
+    :param cost: observed costs for initial function evaluations
```
apaleyes: Good catch!

emukit/examples/gp_bayesian_optimization/optimization_loops.py
```diff
@@ -14,13 +14,15 @@
 
 def create_bayesian_optimization_loop(x_init: np.ndarray, y_init: np.ndarray, parameter_space: ParameterSpace,
                                       acquisition_type: AcquisitionType, model_type: ModelType,
+                                      c_init: np.ndarray = None,
```
apaleyes: Let's be more expressive
```suggestion
                                      cost_init: np.ndarray = None,
```

apaleyes:
Actually one more thing. Can you add a unit test that verifies initial cost is used as expected? Can be as simple as "create bo loop, make sure cost is in the state, done"

emukit/examples/gp_bayesian_optimization/optimization_loops.py
aaronkl: good point

aaronkl:
ok I've updated a simple unit test that check whether the all the initial data is in the loop state

apaleyes:
Looks good, but the build fails for some reason

aaronkl:
hmm wired, I dont understand why it suddenly throws an import error?

aaronkl:
Ok after some digging, the problem seems to be a bug in pytorch [here](https://github.com/pytorch/pytorch/issues/20030)
It seems that they are already working on a PR. In the meantime: `brew install libomp` seems to solve the problem

apaleyes:
Ok, few options:
1. Wait until they release the fix in pytorch
2. Amend our build and, once the bug is fixed, reverse this change
3. Install older version of pytorch that didn't have the bug yet. again reverse once fixed

2 and 3 would be best i guess. if you go down that route, can you please create an issue and reference the PR or whatever we can from PyTorch repo?

aaronkl:
ok let's go for 2 since I don't know after which pytorch version that bug occurred

apaleyes:
So the build has passed, but running times are concerning: over 20 minutes on macos. @aaronkl how long does  this new test run locally for you?

aaronkl:
it's seems that the test for Bohamiann take much longer than necessary. I will send another PR to solve that.

apaleyes:
Much better times now, merging

PR #202, opened 2019-05-17
Extend benchmarking module

aaronkl:
This PR augments the benchmarking functionality of emutkit by:
 - passing a complete loop state, rather than just X and Y, to the outer loop during benchmarking 
 - enable the outer loop to have multiple event handler in order to allow to measure multiple metrics
 - adds a new metric that computes the cumulative cost after each function evaluation
 - update the unit test to the new interface


emukit/core/loop/outer_loop.py
```diff
@@ -51,8 +51,8 @@ def __init__(self, candidate_point_calculator: CandidatePointCalculator,
         if self.loop_state is None:
             self.loop_state = LoopState([])
 
-        self.loop_start_event = EventHandler()
-        self.iteration_end_event = EventHandler()
+        self.loop_start_event = []
```
marpulli: I don't think this is necessary. The `EventHandler` already acts as a list. You can do `self.loop_start_event.append(new_event_fcn)` and doing `self.iteration_end_event(self, self.loop_state)` with call all functions attached to the event. Let me know if I'm missing something though

apaleyes: That is correct, and is the primary reason EventHandler was introduced: to handle multiple event handlers

marpulli:
This looks good apart from the EventHandler stuff and a unit test for the new CumulativeCost metric

emukit/core/loop/outer_loop.py
aaronkl: ah I see, that makes sense

emukit/benchmarking/loop_benchmarking/metrics.py
```diff
@@ -17,7 +17,8 @@ class MeanSquaredErrorMetric(Metric):
     """
     Mean squared error metric stored in loop state metric dictionary with key "mean_squared_error".
     """
-    def __init__(self, x_test: np.ndarray, y_test: np.ndarray, name: str='mean_squared_error'):
+
+    def __init__(self, x_test: np.ndarray, y_test: np.ndarray, name: str = 'mean_squared_error'):
```
apaleyes: actually the one without spaces is what PEP recommends: "Don't use spaces around the = sign when used to indicate a keyword argument"

applies here and below

emukit/benchmarking/loop_benchmarking/metrics.py
```diff
@@ -81,3 +84,33 @@ def reset(self) -> None:
         Resets the start time
         """
         self.start_time = time()
+
+
+class CumulativeCostMetric(Metric):
+    """
+    The result is stored in the "metrics" dictionary in the loop state with the key "cumulative_costs"
```
apaleyes: Can you extend that a bit, with a description of what is understood as "cumulative cost"?

emukit/benchmarking/loop_benchmarking/metrics.py
```diff
@@ -81,3 +84,33 @@ def reset(self) -> None:
...
+    def __init__(self, name: str = 'cumulative_costs'):
+        self.name = name
+        self.cumulative_costs = np.array([0.0])
+        self.last_observed_iter = 0
+
+    def evaluate(self, loop: OuterLoop, loop_state: LoopState) -> np.ndarray:
+        """
+        Accumulates the cost of each function evaluation
+        :param loop: Outer loop
+        :param loop_state: Object containing history of the loop that we add results to
```
apaleyes: Add "returns" section

emukit/benchmarking/loop_benchmarking/benchmarker.py
```diff
@@ -1,5 +1,6 @@
 import logging
 import numpy as np
+from functools import partial
```
apaleyes: Never heard of "partial" before. Thanks, learned something new today!

Issue #206, opened 2019-05-20
implementing random search baseline


aaronkl:
It would be very convenient for developing and benchmarking to have a random search baseline that implements the OuterLoop interface such that, for example, it can easily be plugged into the benchmarking module
A trivial way to do that would be to implement a dummy ModelUpdate and a new CandidatePointSelector which simply returns a random data point. What do you think?

marpulli:
I agree this would be very useful and the way you suggest is sensible. 

It might also be interesting to include some other, potentially stronger, baselines such as Sobol sequences or Latin Hypercubes. The implementation would be a bit more complex however.

PR #208, opened 2019-05-21
Random search

aaronkl:
Implements a simple random search baseline as described in #206 


emukit/examples/gp_bayesian_optimization/random_search.py
```diff
@@ -0,0 +1,74 @@
...
+        self.rd = RandomDesign(parameter_space=parameter_space)
+
+    def compute_next_points(self, loop_state: LoopState, context: dict=None) -> np.ndarray:
+        """
+        :param loop_state: Object that contains current state of the loop
+        :param context: Contains variables to fix through optimization of acquisition function. The dictionary key is
+                        the parameter name and the value is the value to fix the parameter to.
+        :return: (n_points x n_dims) array of next inputs to evaluate the function at
+        """
+        return self.rd.get_samples(1)
```
marpulli: This currently ignores the context. You could do:

```python
if context_manager is not None:
    sample = context_manager.context_free_space.sample_uniform(1)
    sample = context_manager.expand_vector(sample)
else:
    sample = self.parameter_space.sample_uniform(1)
return sample
```

aaronkl: good point, I will add that

emukit/examples/gp_bayesian_optimization/random_search.py
```diff
@@ -0,0 +1,78 @@
...
+
+        if x_init is not None and y_init is not None:
+            loop_state = create_loop_state(x_init, y_init, cost_init)
+        else:
+            loop_state = None
+
+        super().__init__(candidate_point_calculator, model_updaters, loop_state=loop_state)
+
+    def get_results(self):
+        return BayesianOptimizationResults(self.loop_state)
```
apaleyes: That's the only weird thing in the whole PR. RandomSearch has no reason to return BayesOpt results. What if I want to use it as a baseline for quadrature?

integration_tests/emukit/bayesian_optimization/test_random_search.py
```diff
@@ -0,0 +1,14 @@
+import numpy as np
```
apaleyes: Given what I said above, this file may need some splitting/refactoring

emukit/examples/gp_bayesian_optimization/random_search.py
aaronkl: The idea was to provide the same output as BayesOpt for comparisons. However, if we move to the core library I will drop it.

aaronkl:
I like the idea of merging it into the core library, since I could also see some other cases where these functionalities might be useful

emukit/core/loop/model_updaters.py
```diff
@@ -24,6 +24,16 @@ def update(self, loop_state: LoopState) -> None:
         pass
 
 
+class NoopModelUpdate(ModelUpdater):
```
apaleyes:
```suggestion
class NoopModelUpdater(ModelUpdater):
```

emukit/core/loop/model_updaters.py
```diff
@@ -24,6 +24,16 @@ def update(self, loop_state: LoopState) -> None:
...
+    def update(self, loop_state: LoopState) -> None:
+        """
+        Dummy model update, can be used for example for random search
```
apaleyes:
```suggestion
        Dummy model updater that does nothing, can be used for example for random search
```

emukit/benchmarking/loop_benchmarking/random_search.py
```diff
@@ -0,0 +1,36 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import numpy as np
+
+from emukit.core.loop.outer_loop import OuterLoop
```
apaleyes: Inside the library all imports should be relative

tests/emukit/benchmarking/test_random_search.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def test_random_search_with_init_data():
+    np.random.seed(42)
+    branin_fcn, parameter_space = branin_function()
+
+    x_init = parameter_space.sample_uniform(5)
+    y_init = branin_fcn(x_init)
+    cost_init = np.ones([5, 1])
+    print(cost_init.ndim)
```
apaleyes: Remove print, i suppose some debugging leftover?

PR #209, opened 2019-05-24
Fix error in SA notebook.

jejjohnson:
Two Minor Issues:

1) I found a small error where the sensitivity analysis using the Gaussian process regression (GPR) algorithm. The GPR algorithm was not optimized before it was passed through to the MonteCarloSensitivity class. 

2) The plot didn't display the first-order Sobel indexes for the GPR algorithm. The output of the sensitivity class; wrapped around the GPyModelWrapper; returns a dataframe where every entry is an np.ndarray. This causes issues when plotting with the pandas dataframe thus prohibiting the GPR model results from being plotted. I did a simple dictionary comprehension to change the entries from np.ndarray to floats.


notebooks/Emukit-tutorial-sensitivity-montecarlo.ipynb
```diff
@@ -657,9 +658,9 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "Python [conda env:emukit]",
```
apaleyes: These kernel changes don't look right, and they brake the notebook validation tests. Can you revert them (this one, and two lines below)?

jejjohnson: I reverted the changes to the previous kernel names. The updated python version is ok or should I revert that one as well?

Issue #210, opened 2019-05-26
Fixed sample of representer points for entropy search used throughout whole optimization


henrymoss:
Hi 

The provided examples of entropy search and the multi-fidelity entropy search acquisitions currently only sample a single set of representer points, which they then use for the whole optimization.

This set should ideally be re-sampled at the beginning of each BO step (after updating the GP and proposal function), adaptively responding to our observations and focusing computational resources on promising areas.

Using the same set throughout requires a much larger choice of sample size to cover the search space and allow high-precision optimization.

My current workaround is to add a step that does this resampling at the end of each BO step.
```

def resampler(loop,loop_state):
    loop.candidate_point_calculator.acquisition.update_parameters()
loop.iteration_end_event.append(resampler)
```

I think this should be made clearer in the documentation or added automatically within the OuterLoop utility.

javiergonzalezh:
Hi Henri, 

Good catch!  Do you mind doing a pull request with the change? We will review it and merge.

apaleyes:
Hey @henrymoss ! Would you be able to answer Mark's question here? I'll copy it over:

> I believe this problem originates from the fact that update_parameters is not called on the CostAcquisition. Are you using CostAcquisition or acquisition_per_expected_cost in your code for issue #210 ?

henrymoss:
Sorry for the Delay!

I was using the Cost acquisition (as used in https://nbviewer.jupyter.org/github/amzn/emukit/blob/master/notebooks/Emukit-tutorial-multi-fidelity-bayesian-optimization.ipynb). I had just changed the acquisition function.

marpulli:
If I run the following minimal example, the representer points do change after each iteration. Could you give me a minimal working example where you have found that they don't so I can debug the problem please?

```
from emukit.test_functions import forrester_function
from emukit.core.loop.user_function import UserFunctionWrapper
from emukit.model_wrappers import GPyModelWrapper
from emukit.bayesian_optimization.acquisitions import EntropySearch
from emukit.bayesian_optimization.loops.cost_sensitive_bayesian_optimization_loop import CostSensitiveBayesianOptimizationLoop
from emukit.core.acquisition import acquisition_per_expected_cost

import GPy
import numpy as np

# Get user function - cost is just constant 1
target_function, space = forrester_function()
user_func = UserFunctionWrapper(lambda x: (target_function(x), np.array([[1]])), extra_output_names=['cost'])

# Initial data points
[...]

PR #211, opened 2019-05-27
fix to regenerate sample used for entropy search at each optimization step

henrymoss:
Issue 210

If using the entropy search acquisition function, then this needs to be regenerated for each step in the optimization. 


marpulli:
`update_parameters` is already called in the `SequentialPointCalculator` so I don't think think this PR is the way to fix the issue #210. 

I believe this problem originates from the fact that `update_parameters` is not called on the `CostAcquisition`. Are you using `CostAcquisition` or `acquisition_per_expected_cost` in your code for issue #210 ?

apaleyes:
I agree with Mark on this. Let's close the PR and continue investigation in #210

PR #212, opened 2019-05-28
Inequality constraints in parameter space

marpulli:
Add ability to add inequality constraints to parameter space. This includes:
- Adding linear and non-linear inequality constraint objects as well as an IConstraint interface.
- Adding the scipy Trust Region Constrained optimizer that can optimize the acquisition in the presence of inequality constraints
- An `AnchorPointsGenerator` that will get anchor points that satisfy the constraints.


emukit/core/constraints.py
```diff
@@ -0,0 +1,73 @@
...
+        :param lower_bound: Lower bound vector of size (n_constraint,). Can be -np.inf for one sided constraint
+        :param upper_bound: Upper bound vector of size (n_constraint,). Can be np.inf for one sided constraint
+        :param jacobian_fun: Function describing
+        """
+        self.fun = fun
+        self.lower_bound = lower_bound
+        self.upper_bound = upper_bound
+        self.jacobian_fun = jacobian_fun if jacobian_fun else '2-point'
+
+    def evaluate(self, X: np.ndarray) -> np.ndarray:
```
mmahsereci: input should be small x to be more consistent.

emukit/core/constraints.py
```diff
@@ -0,0 +1,73 @@
...
+
+import numpy as np
+
+_log = logging.getLogger(__name__)
+
+
+class IConstraint:
+    def evaluate(self, x: np.ndarray) -> np.ndarray:
+        """
+        :param x: Input locations to evaluate constraint at
```
mmahsereci: shape missing in all evaluate methods.

tests/emukit/core/optimization/test_trust_region_constrained_optimizer.py
```diff
@@ -0,0 +1,44 @@
...
+@pytest.fixture
+def trust_region_constr_linear_constraint():
+    constraints = [LinearInequalityConstraint(np.array([[0, 1]]), np.array([0.5]), np.array([np.inf]))]
+    return OptTrustRegionConstrained([(-1, 1), (-1, 1)], constraints, 1000)
+
+
+def test_trust_region_constrained_no_context(trust_region_constr_linear_constraint, objective, space):
+    x0 = np.array([1, 1])
+    x, f = apply_optimizer(trust_region_constr_linear_constraint, x0, space, objective, None, None, None)
+    print(x)
```
mmahsereci: remove print statement

tests/emukit/core/optimization/test_trust_region_constrained_optimizer.py
```diff
@@ -0,0 +1,44 @@
...
+    assert np.all(np.isclose(x, np.array([0, 0.5])))
+
+
+def test_trust_region_constrained_no_context(trust_region_constr_linear_constraint, objective, space):
+    x0 = np.array([1, 1])
+    x, f = apply_optimizer(trust_region_constr_linear_constraint, x0, space, objective, None, None, None)
+    print(x)
```
mmahsereci: remove print statement

emukit/core/constraints.py
```diff
@@ -0,0 +1,73 @@
...
+                 constraint and zero if the constraint is violated
+        """
+        raise NotImplementedError
+
+
+class LinearInequalityConstraint(IConstraint):
+    """
+    Constraint of the form lower_bound < Ax < upper_bound where the matrix A is called "constraint_matrix"
+    """
+    def __init__(self, constraint_matrix: np.ndarray, lower_bound: np.ndarray, upper_bound: np.ndarray):
```
apaleyes: I think this could use some defaults. Say identity matrix, -infinity, +infinity. As well as validation, that A!=0 and at least one bound is set.

emukit/core/constraints.py
```diff
@@ -0,0 +1,73 @@
...
+    def evaluate(self, x: np.ndarray) -> np.ndarray:
+        """
+        :param x: Input locations to evaluate constraint at
+        :return: Numpy array of shape (n_input,) where an element will be 1 if the corresponding input satisfies the
+                 constraint and zero if the constraint is violated
+        """
+        raise NotImplementedError
+
+
+class LinearInequalityConstraint(IConstraint):
```
apaleyes: Isn't that just NonlinearInequalityConstraint with g(x) = Ax ? you could simplify the code a bit with this

emukit/core/constraints.py
```diff
@@ -0,0 +1,73 @@
...
+
+        :param x: Input locations to evaluate constraint at
+        :return: Numpy array of shape (n_input,) where an element will be 1 if the corresponding input satisfies the
+                 constraint and zero if the constraint is violated
+        """
+        ax = self.constraint_matrix.dot(x)
+        return np.all([ax >= self.lower_bound, ax <= self.upper_bound], axis=0)
+
+
+class NonlinearInequalityConstraint:
```
apaleyes: if g(x) = Ax this becomes linear (see above). So how about just "InequalityConstraint" for the name?

emukit/core/constraints.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    Constraint of the form lower_bound < g(x) < upper_bound
+    """
+    def __init__(self, fun: Callable, lower_bound: np.ndarray, upper_bound: np.ndarray,
```
apaleyes: again, i think defaults and validation for bounds would be good addition here

emukit/core/optimization/anchor_points_generator.py
```diff
@@ -76,9 +76,36 @@ def __init__(self, space: ParameterSpace, acquisition: Acquisition, num_samples=
...
-    def get_anchor_point_scores(self, X) -> np.array:
+    def get_anchor_point_scores(self, X: np.ndarray) -> np.ndarray:
         """
         :param X: The samples at which to evaluate the criterion
         :return:
         """
         return self.acquisition.evaluate(X).flatten()
+
+
+class ConstrainedObjectiveAnchorPointsGenerator(AnchorPointsGenerator):
```
apaleyes: Maybe we don't need a new class here? ObjectiveAnchorPointsGenerator is just ConstrainedObjectiveAnchorPointsGenerator with empty list of constraints, isn't it?

emukit/core/optimization/context_manager.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, space: ParameterSpace, context: Context):
             # Find encoded values of context variable
             param = self.space.get_parameter_by_name(name)
             if hasattr(param, 'encoding'):
-                self.context_values.append(param.encoding.get_encoding(context[name]))
+                self.context_values.append(int(param.encoding.get_encoding(context[name])))
```
apaleyes: Why is this needed?

emukit/core/optimization/gradient_acquisition_optimizer.py
```diff
@@ -67,3 +66,15 @@ def f_df(x):
 
         x_min, fx_min = min(optimized_points, key=lambda t: t[1])
         return x_min, -fx_min
+
+    def _get_anchor_points_generator(self, acquisition):
```
apaleyes: As noted above, we may be able to get rid of this getter

emukit/core/constraints.py
marpulli: I'll add defaults for the bounds and some validation, I'm not keen on identity for constraint_matrix as you should be using the parameter bounds in that case rather than a constraint.

emukit/core/constraints.py
marpulli: The optimizer will treat linear and non-linear constraints differently so it needs to know whether a certain constraint is linear or non-linear and they have to be represented differently. I don't think it's possible to merge the two inequality constraints.

Setting g(x) = Ax is possible but you would get lots of warnings from the optimizer and would be inefficient.

emukit/core/constraints.py
marpulli: Same as above

emukit/core/constraints.py
marpulli: I'll add them

emukit/core/optimization/anchor_points_generator.py
marpulli: Good point, I'll change this

emukit/core/optimization/context_manager.py
marpulli: My bad, it's not. I'll remove.

tests/emukit/core/optimization/test_trust_region_constrained_optimizer.py
marpulli: Fixed

emukit/core/constraints.py
marpulli: Fixed

emukit/core/constraints.py
marpulli: Fixed

tests/emukit/core/optimization/test_trust_region_constrained_optimizer.py
marpulli: Fixed

emukit/core/constraints.py
```diff
@@ -0,0 +1,97 @@
...
+        :param lower_bound: Lower bound vector of size (n_constraint,). Can be -np.inf for one sided constraint
+        :param upper_bound: Upper bound vector of size (n_constraint,). Can be np.inf for one sided constraint
+        """
+        if (lower_bound is None) and (upper_bound is None):
+            raise ValueError('Neither lower nor upper bounds is set, at least one must be specified')
+
+        # Default lower bound to -infinity
+        n_constraints = constraint_matrix.shape[0]
+        if lower_bound is None:
+            lower_bound = np.full([n_constraints], -np.nan)
```
apaleyes: Did you mean `np.inf` here and below?

emukit/core/constraints.py
```diff
@@ -0,0 +1,97 @@
...
+    Constraint of the form lower_bound < Ax < upper_bound where the matrix A is called "constraint_matrix"
+    """
+    def __init__(self, constraint_matrix: np.ndarray, lower_bound: np.ndarray=None, upper_bound: np.ndarray=None):
+        """
+
+        :param constraint_matrix: (n_constraint, n_x_dims) matrix in b_lower < Ax < b_upper
+        :param lower_bound: Lower bound vector of size (n_constraint,). Can be -np.inf for one sided constraint
+        :param upper_bound: Upper bound vector of size (n_constraint,). Can be np.inf for one sided constraint
+        """
+        if (lower_bound is None) and (upper_bound is None):
```
apaleyes: You may also want to make sure they are not +-inf at the same time, maybe after defaults are handled

emukit/core/constraints.py
apaleyes: Ok, I appreciate you need a way to differentiate inequalities. However current approach creates quite a bit of code repetition.

Maybe then we could have

```
class InequalityConstraint(IConstraint):

class LinearIC(InequalityConstraint):

class NonlinearIC(InequalityConstraint):
```

emukit/core/constraints.py
```diff
@@ -0,0 +1,97 @@
...
+    Constraint of the form lower_bound < g(x) < upper_bound
+    """
+    def __init__(self, fun: Callable, lower_bound: np.ndarray, upper_bound: np.ndarray,
+                 jacobian_fun: Optional[Callable]=None):
+        """
+        :param fun: function defining constraint in b_lower < fun(x) < b_upper. Has signature f(x) -> array, shape(m,)
+                    where x is 1d and m is the number of constraints
+        :param lower_bound: Lower bound vector of size (n_constraint,). Can be -np.inf for one sided constraint
+        :param upper_bound: Upper bound vector of size (n_constraint,). Can be np.inf for one sided constraint
+        :param jacobian_fun: Function describing
```
apaleyes: describing...

Seems that is comes directly from scipy, can you add a reference here?

emukit/core/optimization/optimizer.py
```diff
@@ -171,3 +173,70 @@ def f_df_no_context(self, x: np.array) -> Tuple[np.array, np.array]:
...
+
+        :param x0: Initial start point
+        :param f: Function to optimize
+        :param df: Derivative of function to optimize
+        :param f_df: Function returning both value of objective and its gradient
+        :return:
+        """
+        if df is None and f_df is not None:
+            df_1d = lambda x: f_df(x)[1]
+        elif df is not None:
```
apaleyes: Some true magic happening here. What if `df` and `f_df` are both given, looks like this code will overwrite `f_df` then?

emukit/core/constraints.py
apaleyes: Ok, agreement is not to tie this to scipy at all, so just expand the comment and move default string literal value for it to scipy conversion code inside the optimizer

emukit/core/constraints.py
marpulli: Added that validation below

emukit/core/constraints.py
marpulli: This is now done

emukit/core/constraints.py
marpulli: Yep, this has been changed

emukit/core/optimization/gradient_acquisition_optimizer.py
marpulli: Now removed

emukit/core/constraints.py
marpulli: This has been done

marpulli:
I've now addressed all comments and I think it's ready to merge - @apaleyes could you approve if you agree please?

tests/emukit/test_acquisitions.py
```diff
@@ -98,6 +98,10 @@ def integrated_variance_acquisition(gpy_model, continuous_space):
 def squared_correlation_acquisition(vanilla_bq_model):
     return SquaredCorrelation(vanilla_bq_model)
 
+@pytest.fixture
+def probability_of_feasibility(gpy_model):
+    return ProbabilityOfFeasibility(gpy_model)
```
apaleyes: I don't think we have this one merged yet, do we?

marpulli: That's weird - I'm not sure how that got there. I've removed it

PR #215, opened 2019-05-31
Simple GP model to use with intro code

marpulli:
*Issue #, if available:* #140 

*Description of changes:* This is a simple GP model that would help remove GPy as a dependency. It should remove the need for GPy in our simple intro examples.


tests/emukit/examples/simple_gp_model.py
```diff
@@ -0,0 +1,19 @@
...
+def test_simple_gp_model_optimize():
+    x = np.linspace(-1, 1, 3)[:, None]
+    y = x **2
+    gp = SimpleGaussianProcessModel(x, y)
+    gp.optimize()
+
+
+def test_simple_gp_model_predict():
+    x = np.linspace(-1, 1, 3)[:, None]
+    y = x **2
```
apaleyes:
```suggestion
    y = x**2
```

tests/emukit/examples/simple_gp_model.py
```diff
@@ -0,0 +1,19 @@
+import numpy as np
+
+from emukit.examples.simple_gp_model.simple_gp_model import SimpleGaussianProcessModel
+
+
+def test_simple_gp_model_optimize():
+    x = np.linspace(-1, 1, 3)[:, None]
```
apaleyes: x and y definition repeat in both tests, maybe use a fixture for that?

emukit/examples/simple_gp_model/simple_gp_model.py
```diff
@@ -0,0 +1,121 @@
...
+    def predict(self, x_new: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
+        """
+        Predict from model
+
+        :param x_new: (n_points, n_dims) array containing points at which the predictive distributions will be computed
+        :return: Tuple containing two (n_points, 1) arrays representing the mean and variance of the predictive
+                 distribution at the specified input locations
+        """
+        K = self._calc_kernel(self.x)
+        idxs = list(range(self.x.shape[0]))
```
apaleyes: Seems that it can also be done as

    K[idxs, idxs] += np.identity(self.x.shape[0]) * (self.jitter + self.likelihood_variance)

i personally find this a bit more intuitive

emukit/examples/simple_gp_model/simple_gp_model.py
```diff
@@ -0,0 +1,121 @@
...
+        self.kernel_variance = 1
+        self.likelihood_variance = 1
+        self.jitter = 1e-6
+
+    def optimize(self) -> None:
+        """
+        Optimize the three hyperparameters of the model, namely the kernel variance, kernel lengthscale and likelihood
+        variance
+        """
+        def optimize_fcn(x):
```
apaleyes: Would you mind renaming x to something else here? I got really confused between this and model's x

emukit/examples/simple_gp_model/simple_gp_model.py
```diff
@@ -0,0 +1,121 @@
...
+
+        mu = np.dot(tmp.T, tmp2)
+        var = (self.kernel_variance - np.sum(np.square(tmp), axis=0))[:, None]
+        return mu, var
+
+    def _calc_kernel(self, X, X2=None):
+        """
+        Implements an RBF kernel with no ARD
+        """
+        if X2 is None:
```
apaleyes: Looks like this if/else can be re-written as:
```
if X2 is None:
    X2 = X.copy()

X1sq = np.sum(np.square(X), 1)
X2sq = np.sum(np.square(X2), 1)
r2 = -2. * np.dot(X, X2.T) + (X1sq[:, None] + X2sq[None, :])
r2 = np.clip(r2, 0, np.inf)
```

maybe `.copy()` isn't even needed

tests/emukit/examples/simple_gp_model.py
```diff
@@ -0,0 +1,19 @@
...
+    y = x **2
+    gp = SimpleGaussianProcessModel(x, y)
+    gp.optimize()
```
apaleyes: this test case is missing any assertions

tests/emukit/examples/simple_gp_model.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def test_simple_gp_model_predict():
+    x = np.linspace(-1, 1, 3)[:, None]
+    y = x **2
+    gp = SimpleGaussianProcessModel(x, y)
+    mean, var = gp.predict(np.array(5, 1))
+    assert mean.shape == (5, 1)
+    assert var.shape == (5, 1)
```
apaleyes: maybe we can do a bit more than just shape testing here? perhaps assert certain bounds of the result?

mmahsereci:
@javiergonzalezh I agree. In a sense this is a phython-gp-model-wrapper. examples/models were usually tight to specific tasks in the examples folder. The line is a bit blurry though. The conversation fits in well with the conversation we had about re-thinking the examples folder. For now I'd put it in the model wrapper folder since that is where I would go looking for it.

emukit/examples/simple_gp_model/simple_gp_model.py
mmahsereci: There might be a typo.  K += .. is enough here I guess with your suggestion.

apaleyes: Right, thanks for the correction Maren

emukit/examples/simple_gp_model/simple_gp_model.py
```diff
@@ -0,0 +1,121 @@
...
+        K_xs = self._calc_kernel(self.x, x_new)
+
+        tmp = scipy.linalg.solve_triangular(L, K_xs, lower=True)
+        tmp2 = scipy.linalg.solve_triangular(L, self.y, lower=True)
+
+        mu = np.dot(tmp.T, tmp2)
+        var = (self.kernel_variance - np.sum(np.square(tmp), axis=0))[:, None]
+        return mu, var
+
+    def _calc_kernel(self, X, X2=None):
```
mmahsereci: the RBF kernel is usually define with a 0.5 in the exponent, i,e. exp(-0.5 * scaled_diff)

emukit/examples/simple_gp_model/simple_gp_model.py
```diff
@@ -0,0 +1,121 @@
...
+        idxs = list(range(self.x.shape[0]))
+        K[idxs, idxs] += self.jitter + self.likelihood_variance
+        L = np.linalg.cholesky(K)
+
+        K_xs = self._calc_kernel(self.x, x_new)
+
+        tmp = scipy.linalg.solve_triangular(L, K_xs, lower=True)
+        tmp2 = scipy.linalg.solve_triangular(L, self.y, lower=True)
+
+        mu = np.dot(tmp.T, tmp2)
```
mmahsereci: Let's call those mean and variance instead of mu and var.

emukit/examples/simple_gp_model/simple_gp_model.py
mmahsereci: type annotation missing, and docstring params and return missing

emukit/examples/simple_gp_model/simple_gp_model.py
```diff
@@ -0,0 +1,121 @@
...
+            X1sq = np.sum(np.square(X), 1)
+            X2sq = np.sum(np.square(X2), 1)
+            r2 = -2. * np.dot(X, X2.T) + (X1sq[:, None] + X2sq[None, :])
+            r2 = np.clip(r2, 0, np.inf)
+
+        return self.kernel_variance * np.exp(-r2 / self.lengthscale ** 2)
+
+    def _negative_marginal_log_likelihood(self) -> float:
+        """
+        Negative marginal log likelihood
```
mmahsereci: return in docstring missing

emukit/examples/simple_gp_model/simple_gp_model.py
```diff
@@ -0,0 +1,121 @@
...
+        L = np.linalg.cholesky(K)
+
+        # Log determinant of the covariance matrix
+        log_det = 2. * np.sum(np.log(np.diag(L)))
+
+        # calculate y^T K^{-1} y using the cholesky of K
+        tmp = scipy.linalg.solve_triangular(L, self.y, lower=True)
+        alpha = scipy.linalg.solve_triangular(L.T, tmp, lower=False)
+        log_2_pi = np.log(2 * np.pi)
+        return -0.5 * (-self.y.size * log_2_pi - self.y.shape[1] * log_det - np.sum(alpha * self.y))
```
mmahsereci: do you mean self.x.size here instead of self.y.size?

emukit/examples/simple_gp_model/simple_gp_model.py
marpulli: Thanks, this is now done

emukit/examples/simple_gp_model/simple_gp_model.py
marpulli: Yep, changed to `log_hyper_parameters`

emukit/examples/simple_gp_model/simple_gp_model.py
marpulli: Done

tests/emukit/examples/simple_gp_model.py
marpulli: I've redone the tests

tests/emukit/examples/simple_gp_model.py
marpulli: There is now a test that tests the results are close to the training data

emukit/examples/simple_gp_model/simple_gp_model.py
marpulli: Thanks, I've changed this

emukit/examples/simple_gp_model/simple_gp_model.py
marpulli: Done

emukit/examples/simple_gp_model/simple_gp_model.py
marpulli: Added the docstrings now

emukit/examples/simple_gp_model/simple_gp_model.py
marpulli: Added a return type

emukit/examples/simple_gp_model/simple_gp_model.py
marpulli: I think it should be y.size as the nmll only depends on the dimensionality of the output rather than input

mmahsereci: yes, totally. Not sure what I was thinking o_0

marpulli:
I've now moved the model to model_wrappers. It's not truly just a wrapper - but  I think this is still the best place for now

apaleyes:
@marpulli sure, for now that will do, but i guess we will want to revisit this, just because, as you say, this is not a wrapper

PR #217, opened 2019-06-03
Bayesian Optimization with unknown constraint

polivucci:
*Issue #, if available:*

*Description of changes:*


emukit/bayesian_optimization/acquisitions/probability_of_improvement.py
```diff
@@ -67,3 +67,56 @@ def evaluate_with_gradients(self, x: np.ndarray) -> Tuple:
...
+
+    def evaluate(self, x: np.ndarray) -> np.ndarray:
+        """
+        Computes the probability of improving over the current best
+
+        :param x: points where the acquisition is evaluated.
+        """
+        mean, variance = self.model.predict(x)
+        standard_deviation = np.sqrt(variance)
+        _, cdf, _ = get_quantiles(self.jitter, 0, mean, standard_deviation)
```
marpulli: We have removed the get_quantiles import from this file so we don't have to rely on GPyOpt so this line won't run! Could you look at the code for Probability of improvement acquisition and try to factor out the relevant bits so you can reuse it here?

emukit/bayesian_optimization/acquisitions/probability_of_improvement.py
```diff
@@ -67,3 +67,56 @@ def evaluate_with_gradients(self, x: np.ndarray) -> Tuple:
     @property
     def has_gradients(self):
         return isinstance(self.model, IDifferentiable)
+    
+    
+    
+class ProbabilityOfFeasibility(Acquisition):
```
marpulli: Could you add a test for this please? The tests are already written you just need to add a few things in `test_acquisitions.py`:
1. Import this acquisition.
2. A fixture that creates this acquisition
```
@pytest.fixture
def probability_of_feasibility_acquisition(gpy_model):
    return ProbabilityOfFeasibility(gpy_model)
```
3. Add this to the `acquisition_tests` list in the same file:
`acquisition_test_tuple('probability_of_feasibility_acquisition', True, default_grad_tol)`

polivucci: Hi, I have updated test_acquisitions.py following the instructions

emukit/bayesian_optimization/acquisitions/probability_of_improvement.py
```diff
@@ -67,3 +67,56 @@ def evaluate_with_gradients(self, x: np.ndarray) -> Tuple:
     @property
     def has_gradients(self):
         return isinstance(self.model, IDifferentiable)
+    
+    
+
```
marpulli: I think the acquisition should also go in it's own file `probability_of_feasibility.py`

emukit/bayesian_optimization/acquisitions/constrained_bayesian_optimization.py
```diff
@@ -0,0 +1,68 @@
...
+from ...bayesian_optimization.acquisitions import ExpectedImprovement, ProbabilityOfFeasibility
+from ...core.acquisition import Acquisition
+from ...core.interfaces import IModel
+from ...core.loop import FixedIntervalUpdater, OuterLoop, SequentialPointCalculator
+from ...core.loop.loop_state import create_loop_state
+from ...core.optimization import AcquisitionOptimizer
+from ...core.parameter_space import ParameterSpace
+
+
+class ConstrainedBayesianOptimizationLoop(OuterLoop):
```
marpulli: Can we move this to `emukit.bayesian_optimization.loops`? @apaleyes do you agree? Or should it go in examples?

emukit/bayesian_optimization/acquisitions/constrained_bayesian_optimization.py
```diff
@@ -0,0 +1,68 @@
...
+
+        Michael A. Gelbart, Jasper Snoek, and Ryan P. Adams,
+        Bayesian Optimization with Unknown Constraints,
+        https://arxiv.org/pdf/1403.5607.pdf
+
+        :param space: Input space where the optimization is carried out.
+        :param model_objective: The model that approximates the underlying objective function
+        :param model_constraint: The model that approximates the unknown constraints
+        :param acquisition: The acquisition function that will be used to collect new points (default, CEI).
+        :param update_interval:  Number of iterations between optimization of model hyper-parameters. Defaults to 1.
```
mmahsereci: batch_size param missing in docstring

emukit/bayesian_optimization/acquisitions/constrained_bayesian_optimization.py
```diff
@@ -0,0 +1,68 @@
...
+        if not np.all(np.isclose(model_objective.X, model_constraint.X)):
+            raise ValueError('Emukit currently only supports identical '
+                             'training inputs for the constrained and objective model')
+
+        if acquisition is None:
+            acquisition_objective = ExpectedImprovement(model_objective)
+            acquisition_constraint = ProbabilityOfFeasibility(model_constraint)
+            acquisition_constrained = acquisition_objective * acquisition_constraint
+
+
```
mmahsereci: remove blank line

emukit/bayesian_optimization/acquisitions/constrained_bayesian_optimization.py
```diff
@@ -0,0 +1,68 @@
...
+        model_updater_objective = FixedIntervalUpdater(model_objective, update_interval)
+        model_updater_constraint = FixedIntervalUpdater(model_constraint, update_interval, lambda state: state.cost)
+
+        acquisition_optimizer = AcquisitionOptimizer(space)
+        if batch_size == 1:
+            candidate_point_calculator = SequentialPointCalculator(acquisition, acquisition_optimizer)
+        else:
+            if not isinstance(model, IDifferentiable):
```
mmahsereci: model does not seem to be defined. this may indicate a useful test.

emukit/bayesian_optimization/acquisitions/constrained_bayesian_optimization.py
```diff
@@ -0,0 +1,68 @@
...
+                raise ValueError('Model must implement ' + str(IDifferentiable) +
+                                 ' for use with Local Penalization batch method.')
+            log_acquisition = LogAcquisition(acquisition)
+            candidate_point_calculator = LocalPenalizationPointCalculator(log_acquisition, acquisition_optimizer, model,
```
mmahsereci: same here, model not defined.

emukit/bayesian_optimization/acquisitions/probability_of_improvement.py
```diff
@@ -67,3 +67,56 @@ def evaluate_with_gradients(self, x: np.ndarray) -> Tuple:
...
+        :param model: The underlying model that provides the predictive mean and variance for the given test points
+        :param jitter: Jitter to balance exploration / exploitation
+        """
+        self.model = model
+        self.jitter = jitter
+
+    def evaluate(self, x: np.ndarray) -> np.ndarray:
+        """
+        Computes the probability of improving over the current best
```
mmahsereci: Check this description, please. I assume it is copy paste from the probability of improvement. Please update it necessary.

emukit/bayesian_optimization/acquisitions/probability_of_improvement.py
```diff
@@ -67,3 +67,56 @@ def evaluate_with_gradients(self, x: np.ndarray) -> Tuple:
...
+
+        :param x: points where the acquisition is evaluated.
```
mmahsereci: add the shape please, e.g., (num_points, n_dims)

also the return is missing in the docstring

emukit/bayesian_optimization/acquisitions/probability_of_improvement.py
```diff
@@ -67,3 +67,56 @@ def evaluate_with_gradients(self, x: np.ndarray) -> Tuple:
...
+        """
+        mean, variance = self.model.predict(x)
+        standard_deviation = np.sqrt(variance)
+        _, cdf, _ = get_quantiles(self.jitter, 0, mean, standard_deviation)
+
+        return cdf
+
+    def evaluate_with_gradients(self, x: np.ndarray) -> Tuple:
+        """
+        Computes the  probability of improving over the current best and its derivative
```
mmahsereci: same here, docstring might need to be updated to fit the probability of feasibility acquisition.

emukit/bayesian_optimization/acquisitions/probability_of_improvement.py
```diff
@@ -67,3 +67,56 @@ def evaluate_with_gradients(self, x: np.ndarray) -> Tuple:
...
+
+        :param x: points where the acquisition is evaluated.
```
mmahsereci: add shape here, too, please. (num_points, n_dim)

also the return is missing in the docstring

emukit/bayesian_optimization/acquisitions/constrained_bayesian_optimization.py
mmahsereci: it seems generic enough to go into emukit.bayesian_optimization.loops?

apaleyes: Absolutely. I am just not sure about the name. Is it too generic for what this loop does? How would one call bayes opt loop when constraints are known, like in #212?

emukit/bayesian_optimization/acquisitions/constrained_bayesian_optimization.py
```diff
@@ -0,0 +1,68 @@
...
+from ...core.loop import FixedIntervalUpdater, OuterLoop, SequentialPointCalculator
+from ...core.loop.loop_state import create_loop_state
+from ...core.optimization import AcquisitionOptimizer
+from ...core.parameter_space import ParameterSpace
+
+
+class ConstrainedBayesianOptimizationLoop(OuterLoop):
+
+    def __init__(self, space: ParameterSpace, 
+                 model_objective: IModel,
```
apaleyes: ExpectedImprovement needs IDifferentiable, so this should be `Union[IModel, IDifferentiable]`

emukit/bayesian_optimization/acquisitions/constrained_bayesian_optimization.py
```diff
@@ -0,0 +1,68 @@
...
+from ...core.loop.loop_state import create_loop_state
+from ...core.optimization import AcquisitionOptimizer
+from ...core.parameter_space import ParameterSpace
+
+
+class ConstrainedBayesianOptimizationLoop(OuterLoop):
+
+    def __init__(self, space: ParameterSpace, 
+                 model_objective: IModel, 
+                 model_constraint: IModel,
```
apaleyes: Ditto: `Union[IModel, IDifferentiable]`

emukit/bayesian_optimization/acquisitions/constrained_bayesian_optimization.py
```diff
@@ -0,0 +1,68 @@
...
+            candidate_point_calculator = SequentialPointCalculator(acquisition, acquisition_optimizer)
+        else:
+            if not isinstance(model, IDifferentiable):
+                raise ValueError('Model must implement ' + str(IDifferentiable) +
+                                 ' for use with Local Penalization batch method.')
+            log_acquisition = LogAcquisition(acquisition)
+            candidate_point_calculator = LocalPenalizationPointCalculator(log_acquisition, acquisition_optimizer, model,
+                                                                          space, batch_size)
+
+        loop_state = create_loop_state(model_objective.X, model_objective.Y, model_constraint.Y)
```
apaleyes: So `model_constraint.Y` is treated as cost? I don't know if that's right or wrong, just clarifying for myself.

emukit/bayesian_optimization/acquisitions/probability_of_improvement.py
apaleyes: agreed

emukit/bayesian_optimization/acquisitions/probability_of_improvement.py
```diff
@@ -67,3 +67,56 @@ def evaluate_with_gradients(self, x: np.ndarray) -> Tuple:
     @property
     def has_gradients(self):
         return isinstance(self.model, IDifferentiable)
+    
+    
+    
+class ProbabilityOfFeasibility(Acquisition):
+
+    def __init__(self, model: Union[IModel, IDifferentiable], jitter: np.float64 = np.float64(0)) -> None:
```
apaleyes: Unless there is a specific reason to use `np.float64`, let's just use python's `float`

emukit/bayesian_optimization/acquisitions/probability_of_improvement.py
polivucci: I edited the description

emukit/bayesian_optimization/acquisitions/constrained_bayesian_optimization.py
marpulli: I would remove the check - either `LocalPenalizationPointCalculator` needs a model to be differentiable - and so should check this itself, or the model doesn't actually need to be differentiable.

But I agree with @mmahsereci , I would add two integration tests using this loop, one with batch size = 1, one with batch size =2. Checkout `test_single_objective_bayesian_optimization.py` for an example integration test you can copy.

emukit/bayesian_optimization/acquisitions/constrained_bayesian_optimization.py
polivucci: I used cost because it was already available in loop_state,
otherwise I think I should have defined a new variable called e.g. y_constraint
Do you have any suggestion on this?

apaleyes: I do have an idea, but that may be a bit more work. I think it is ok for loops to have their specific state classes. So here you would create ConstrainedLoopState, which would define additional fields you need.

In the hindsight this is probably how cost should have been implemented.

I'd also like to know what others think of this approach. @marpulli @mmahsereci thoughts?

emukit/bayesian_optimization/acquisitions/constrained_bayesian_optimization.py
mmahsereci: What about UnknownConstraintsBayesianOptimizationLoop?

emukit/bayesian_optimization/acquisitions/probability_of_improvement.py
mmahsereci: The ProbabilityOfImprovement declares np.float64, too; this may be a good opportunity to change that as well to float.

emukit/bayesian_optimization/acquisitions/constrained_bayesian_optimization.py
polivucci: what about `model_constraint` ?
do we need its gradients when optimising the acquisition that is the product of 
the two acquisitions?

apaleyes: I am just following the interfaces here. `model_constraint` goes into the new acquisition which delcares that it needs gradients. Maybe it in fact does not need gradients, in which case you should remove them from that acquisition

emukit/bayesian_optimization/acquisitions/constrained_bayesian_optimization.py
polivucci: why is `IDifferentiable` needed for model_objective only?
The final acquisition is acquisition_constraint*acquisition_objective

emukit/bayesian_optimization/acquisitions/constrained_bayesian_optimization.py
polivucci: I have a question on this.
I have modified the code and now `model_objective` is passed to  `LocalPenalizationPointCalculator`.
This means that the local penalizers are computed without regard for the constraints and batch_acquisition = EI * PoF * local_penalizer_1 * ...
I can't say if this naive approach is correct.
Should I move this question to the Issues?

emukit/bayesian_optimization/acquisitions/constrained_bayesian_optimization.py
mmahsereci: I may or may not have another example where a specific loop state would be quite handy @apaleyes. So I like that idea better. (generally speaking)

tests/emukit/test_acquisitions.py
```diff
@@ -37,7 +38,8 @@
                      acquisition_test_tuple('multi_source_entropy_search_acquisition', False, np.nan),
                      acquisition_test_tuple('integrated_variance_acquisition', False, np.nan),
                      acquisition_test_tuple('integrated_expected_improvement_acquisition', True, default_grad_tol),
-                     acquisition_test_tuple('integrated_probability_of_improvement_acquisition', False, np.nan)]
+                     acquisition_test_tuple('integrated_probability_of_improvement_acquisition', False, np.nan)
```
marpulli: This missing comma is causing the tests to fail I think:
```suggestion
                     acquisition_test_tuple('integrated_probability_of_improvement_acquisition', False, np.nan),
```

polivucci: many thanks, 
I am very busy on another project at the moment,
I'll have more time to work on the code from tomorrow
cheers
PO

emukit/bayesian_optimization/loops/unknown_constraint_bayesian_optimization_loop.py
```diff
@@ -0,0 +1,68 @@
...
+        :param acquisition: The acquisition function that will be used to collect new points (default, CEI).
+        :param update_interval:  Number of iterations between optimization of model hyper-parameters. Defaults to 1.
+        :param batch_size: How many points to evaluate in one iteration of the optimization loop. Defaults to 1.
+        """
+
+        if not np.all(np.isclose(model_objective.X, model_constraint.X)):
+            raise ValueError('Emukit currently only supports identical '
+                             'training inputs for the constrained and objective model')
+
+        if acquisition is None:
```
apaleyes: if `acquisition` is not None, none of these below till get assigned. seems wrong

emukit/core/loop/constrained_loop_state.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+from typing import List
+
+import numpy as np
+
+from .user_function_result import UserFunctionResult
+
+
+class UnknownConstraintLoopState(object):
```
apaleyes: This is how I thought this will work, thanks!

emukit/core/loop/constrained_loop_state.py
```diff
@@ -0,0 +1,73 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
apaleyes: Let's keep this file next to the loop file for now, not in "core" package

emukit/bayesian_optimization/acquisitions/constrained_bayesian_optimization.py
mmahsereci: @javiergonzalezh, do you have a good intuition about this? I am not entirely sure.

emukit/bayesian_optimization/loops/unknown_constraint_bayesian_optimization_loop.py
marpulli: I would allow the user to pass in a different `acquisition_objective` but not the `acquisition_constrained` for now, like:
```
if acquisition_objective is None:
    acquisition_objective = ExpectedImprovement(model_objective)

acquisition_constraint = ProbabilityOfFeasibility(model_constraint)
acquisition_constrained = acquisition_objective * acquisition_constraint
```

apaleyes: makes sense!

polivucci:
Thank you, sorry for the delay I was busy on other projects.
Also looking forward to @javiergonzalezh 's opinion on the local penalization issue.

polivucci:
I defined the new loops but then I stopped short of defining a new `UserFunctionResult`.
As it is now, the code still runs with the `cost` variable and the new loops are not used.
Should I remove the new loops?

marpulli:
I just added a test and made this use the latest functionality to pass around arbitrary function outputs (in this case, the constraint value). Once the tests run, I'll merge.

PR #221, opened 2019-06-07
fix bug in interface of single_objective_bayesian_optimisation example

DavidJanz:
*Description of changes:*

single_objective_bayesian_optimisation.py, used within the tutorial examples, seems to have a call to an attribute that no longer exists (perhaps part of old interface?). Removing that call appears to lead to the desired behaviour.


DavidJanz:
@apaleyes  done

PR #223, opened 2019-06-14
Fabolas

aaronkl:
Implements Fabolas as described in https://projecteuclid.org/download/pdfview_1/euclid.ejs/1513306864


emukit/examples/fabolas/continuous_fidelity_entropy_search.py
```diff
@@ -0,0 +1,80 @@
+from typing import Union
+
+import numpy as np
+
+from emukit.bayesian_optimization.acquisitions import ExpectedImprovement
```
apaleyes: Some of these are relative, some absolute imports. I don't mind any inside examples folder, but you should probably pick one.

emukit/examples/fabolas/example.py
```diff
@@ -0,0 +1,22 @@
+from hpolib.benchmarks.surrogates.svm import SurrogateSVM
```
apaleyes: We don't have a dependecy on hpolib, and we shouldn't, so what we do in such cases is wrap this import in try/except and print a friendly message, something like "hpolib is required for this example, run `pip install hpolib`" or something of that sort

emukit/examples/fabolas/continuous_fidelity_entropy_search.py
aaronkl: ok switched all to absolute imports

aaronkl:
Hi Javier, 
actually I wanted to write a notebook that explains how to use Emukit for hyperparameter optimization in general and also shows how Fabolas works. But maybe it would make sense to split that up into two different notebooks?

apaleyes:
@aaronkl we have a couple of bayes opt tutorials, but none of them specifically for HPO, so that makes sense. The only thing is that I am not sure if Fabolas is an obvious first choice for an HPO tutorial

javiergonzalezh:
@aaronkl @apaleyes  having a tutorial about Emukit for HPO would be great. Maybe that notebook can have the flavor of  a small benchmark and include Fabolas in the comparison. Having a second one to explain how Fabolas work, with some context for the method, will be also useful to illustrate some details of approach.

aaronkl:
@javiergonzalezh @apaleyes  yes that makes sense. Should I add that notebook to this PR or create a new PR for that?

apaleyes:
@aaronkl oh no, definitely in a separate PR. i'll merge this one

PR #226, opened 2019-06-24
Update to version 0.4.6

apaleyes:
*Issue #, if available:* #216 

*Description of changes:* The Latin bug in installation is hurting emukit installation, so we need a new version with a fix out.


apaleyes:
As an example, I've run into it here: https://github.com/amzn/xfer/pull/60

PR #232, opened 2019-07-10
adding exp design tutorial

mmahsereci:
*Issue #, if available:* #155

*Description of changes:*

Notebook for experimental design. I may not have found it, but there seems to be no pre-wrapped loop yet that also initialized the model. If this is the case I may add that later as well in the examples. 

Also citations for the ExpDesign acquisition functions are welcome.


notebooks/Emukit-tutorial-experimental-design-introduction.ipynb
```diff
@@ -0,0 +1,622 @@
+{
```
marpulli: <p>This plot is nice, but I woudn't scale the acquisitions just by the maximum value and not shift them. I found it confusing that the IVR acquisition was exactly zero on the right hand side. </p><p><br></p><p>The other thing I found confusing is that the IVR acquisition is not zero where we have already observed point. This is because of the x_plot vector doesn't contain the exact values of the current datapoints, if you change the number of point in x_plot from 200 to 201, it will go through the datapoints and you see the acquisition drop to zero. </p><p><br></p><p>This highlights an interesting property of the IVR vs US: points close to existing training point will give a US acquisition value of ~ 0, whereas the IVR will not be zero. The IVR acquisition recognises that an observation close to, but not exactly on, an existing point will give us some gradient information and therefore some non-local information about f.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/amzn/emukit/pull/232/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

marpulli: <p>I think it should be "ExpDesign" instead of "Bayesian optimization"</p><p><br></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/amzn/emukit/pull/232/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='33'/>

mmahsereci: changed. Also that paragraph is abit of a lie because we do not have a wrapped loop yet that also instantiates the model. I just used the exp design loop.

mmahsereci: I included the observed points in the plot range. Now min(a) is 0 and it would not be shifted, but I removed the shifting anyways since it is a bit confusing. Now the discontinuities are visible.

I had thought of writing about the points you raise as well, but then I left them out because this is just an intro tutorial about how to use emukit, and it might take the focus away from that (and we don't go in depth in the other tutorial either). I hope it is somehow clear what happens from the description of the acquisition functions.

Do you know of generally accepted citations for US and IVR for ExpDesign?

marpulli: I would choose this for the IVAR (they call is IMSE) https://projecteuclid.org/euclid.ss/1177012413. I'm not sure about US.

PR #237, opened 2019-08-06
Fix intergrating hyper-parameters

marpulli:
Integrating hyper-parameters didn't work with multi-output models or acquisitions without gradients.


emukit/core/acquisition/integrated_acquisition.py
```diff
@@ -3,7 +3,7 @@
 import numpy as np
 
 from emukit.core.acquisition import Acquisition
-from emukit.core.interfaces import IModel, IPriorHyperparameters, IDifferentiable
```
apaleyes: Isn't it actually used? I mean, there is definitely a call of `model.evaluate_with_gradients` below. I'd suggest to add this interface to model type hint

emukit/core/acquisition/integrated_acquisition.py
```diff
@@ -47,18 +50,21 @@ def evaluate_with_gradients(self, x: np.ndarray) -> Tuple:
...
 
-        return acquisition_value / self.n_samples, dacquisition_dx / self.n_samples
+    def update_parameters(self):
+        self.samples = self.model.generate_hyperparameters_samples(self.n_samples)
 
     @property
     def has_gradients(self) -> bool:
         """Returns that this acquisition has gradients"""
-        return isinstance(self.model, IDifferentiable)
+        return self._has_gradients
```
apaleyes: Let's cover this with a unit test

emukit/model_wrappers/gpy_model_wrappers.py
```diff
@@ -241,3 +240,33 @@ def get_covariance_between_points(self, X1: np.ndarray, X2: np.ndarray) -> np.nd
...
+        samples = hmc.sample(num_samples = n_burnin + n_samples * subsample_interval, hmc_iters = leapfrog_steps)
+        hmc_samples = samples[n_burnin::subsample_interval]
+        return hmc_samples
+
+    def fix_model_hyperparameters(self, sample_hyperparameters: np.ndarray) -> None:
+        """
+        Fix model hyperparameters
+
+        """
+        if self.gpy_model._fixes_ is None:
```
apaleyes: Is there a way in gpy to do this without calling private stuff like `_fixes_` and `_trigger_params_changed`?

emukit/core/acquisition/integrated_acquisition.py
marpulli: I have always been confused how we should do type hints when some interfaces are optional to implement. The type hint should say "Must implement IModel and optionally implement IDifferentiable". We usually do `Union[IModel, IDifferentialbe]` which means "must implement both IModel and IDifferentialbe".  You get warnings from the type checking system if you pass in a model that doesn't implement gradients with the current type hints.

emukit/core/acquisition/integrated_acquisition.py
marpulli: Done

emukit/model_wrappers/gpy_model_wrappers.py
marpulli: I'll investigate, this is the way the other model wrapper does it though

apaleyes: Following our offline conversation, can you please create an issue to remind us to look deeper into this GPy magic? We may even consider contributing interface changes to GPy if necessary.

apaleyes:
This got broken after #241, and I've submitted a PR with a fix already: #244

PR #247, opened 2019-10-03
gaussian measure for bq

mmahsereci:
*Issue #130 , if available:* 

*Description of changes:*
- added gaussian integration measure
- added q-rbf kernel that integrates agains gauss measure
- deleted integration agains uniform measure *after* the loop. This was not nice in the first place. We can add this as another q-kernel. but I did not want to overload this PR as it is already cluttered.
- adapted tutorial notebook accordingly
- added some tests

Tests pass.


emukit/examples/vanilla_bayesian_quadrature_with_rbf/vanilla_bq_loop_with_rbf.py
```diff
@@ -4,44 +4,50 @@
...
-        D_bounds = len(integral_bounds)
-        input_dim = X.shape[1]
-        raise ValueError("number of integral bounds " + str(D_bounds) + " provided does not match the input dimension "
-                                                                        + str(input_dim) + ".")
+    if integral_bounds is not None:
+        if not len(integral_bounds) == X.shape[1]:
+            D_bounds = len(integral_bounds)
+            input_dim = X.shape[1]
+            raise ValueError("number of integral bounds ", D_bounds, " provided does not match the input "
+                             "dimension ", input_dim, ".")
```
apaleyes: I am entirely sure what way of constructing the error message does. Is the assumption here that all values will be concatenated? 

Even if it is, a clearer way would be to use `"".format` or string interpolation

There are many places in this CR where the comment applies

emukit/examples/vanilla_bayesian_quadrature_with_rbf/vanilla_bq_loop_with_rbf.py
```diff
@@ -4,44 +4,50 @@
...
     :return: The vanilla BQ loop
     """
 
-    if not len(integral_bounds) == X.shape[1]:
-        D_bounds = len(integral_bounds)
-        input_dim = X.shape[1]
-        raise ValueError("number of integral bounds " + str(D_bounds) + " provided does not match the input dimension "
-                                                                        + str(input_dim) + ".")
+    if integral_bounds is not None:
+        if not len(integral_bounds) == X.shape[1]:
```
apaleyes: if ... and ...

emukit/model_wrappers/gpy_quadrature_wrappers.py
```diff
@@ -3,20 +3,22 @@
...
-from emukit.quadrature.kernels.quadrature_rbf import QuadratureRBF
+from emukit.quadrature.kernels.quadrature_rbf import QuadratureRBF, QuadratureRBFIsoGaussMeasure, QuadratureRBFnoMeasure
+from emukit.quadrature.kernels.integration_measures import IntegrationMeasure, IsotropicGaussianMeasure, UniformMeasure
 
 
 class BaseGaussianProcessGPy(IBaseGaussianProcess):
     """
     Wrapper for GPy GPRegression
-    An instance of this can be passed as 'base_gp' to an ApproximateWarpedGPSurrogate object.
+    An instance of this can be passed as 'base_gp' to an WarpedBayesianQuadratureModel object.
```
apaleyes:
```suggestion
    An instance of this can be passed as 'base_gp' to a WarpedBayesianQuadratureModel object.
```

emukit/model_wrappers/gpy_quadrature_wrappers.py
```diff
@@ -162,26 +164,57 @@ def dKdiag_dx(self, x: np.ndarray) -> np.ndarray:
         return np.zeros((input_dim, num_points))
 
 
-def convert_gpy_model_to_emukit_model(gpy_model, integral_bounds: List[Tuple[float, float]], integral_name: str = '') \
+def convert_gpy_model_to_emukit_model(gpy_model, integral_bounds: Union[None, List[Tuple[float, float]]],
```
apaleyes: Since the type of the model is known, feel free to use it as a type hint here.

Also, semantically "create_" would be more correct than "convert_to_" - wrapping is not the same as converting

emukit/model_wrappers/gpy_quadrature_wrappers.py
```diff
@@ -162,26 +164,57 @@ def dKdiag_dx(self, x: np.ndarray) -> np.ndarray:
...
     :param integral_name: the (variable) name(s) of the integral
 
-    :return: emukit model for quadrature witg GPy backend (IBaseGaussianProcessGPy)
+    :return: emukit model for quadrature with GPy backend (IBaseGaussianProcessGPy)
     """
 
-    # wrap standard kernel and get quadrature kernel
-    if gpy_model.kern.name is 'rbf':
+    # wrap standard kernel
+    if gpy_model.kern.name == 'rbf':
```
apaleyes: is `ininstace` possible here?

emukit/quadrature/interfaces/base_gp.py
```diff
@@ -49,6 +49,7 @@ def gram_chol(self) -> np.ndarray:
 
         :return: a lower triangular cholesky of G(X, X)
         """
+        # Todo: is it too much to impose the cholesky? maybe inverse would be better
```
apaleyes: Can you raise an issue explaining this TODO in more detail, in case someone wants to pick it up?

emukit/quadrature/kernels/integration_measures.py
```diff
@@ -0,0 +1,147 @@
...
+    def __init__(self, name: str):
+        """
+        :param name: Name of the integration measure
+        """
+        self.name = name
+
+    def compute_density(self, x: np.ndarray) -> np.ndarray:
+        """
+        Computes the density at point x
+        :param x: points at which density is computes, shape (num_points, dim)
```
apaleyes:
```suggestion
        :param x: points at which density is computed, shape (num_points, dim)
```

emukit/quadrature/kernels/integration_measures.py
```diff
@@ -0,0 +1,147 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+
+import numpy as np
+from typing import Tuple, List
+
+
+class IntegrationMeasure:
+    """An integration measure"""
```
apaleyes:
```suggestion
    """An abstract class for integration measure"""
```

emukit/quadrature/kernels/integration_measures.py
```diff
@@ -0,0 +1,147 @@
...
+    def __init__(self, bounds: List[Tuple[float, float]]):
+        """
+        :param bounds: List of D tuples, where D is the dimensionality of the domain and the tuples contain the lower
+        and upper bounds of the box defining the uniform measure i.e., [(lb_1, ub_1), (lb_2, ub_2), ..., (lb_D, ub_D)]
+        """
+        super().__init__('UniformMeasure')
+
+        # checks if lower bounds are smaller than upper bounds.
+        for bounds_d in bounds:
+            lb_d, ub_d = bounds_d
```
apaleyes: `for (lb_d, ub_d) in bounds:`

emukit/quadrature/kernels/integration_measures.py
```diff
@@ -0,0 +1,147 @@
...
+
+        self.bounds = bounds
+        # uniform measure has constant density which is computed here.
+        self.density = self._compute_density()
+
+    def _compute_density(self) -> float:
+        differences = np.array([x[1] - x[0] for x in self.bounds])
+        volume = np.prod(differences)
+
+        if not volume > 0:
```
apaleyes:
```suggestion
        if volume <= 0:
```

emukit/quadrature/kernels/integration_measures.py
```diff
@@ -0,0 +1,147 @@
...
+            raise NumericalPrecisionError("Domain volume of uniform measure is not positive. Its value is "
+                                          + str(volume) + ". It might be numerical problems...")
```
apaleyes: "It might be numerical problems..." - is it a valid sentence? Does it imply some issues with the input, or that the user may experience issues going forward?

emukit/quadrature/kernels/integration_measures.py
```diff
@@ -0,0 +1,147 @@
...
+        return np.float(1. / volume)
```
apaleyes: Is there any reason to do `np.float` here? The result will already be a float

emukit/model_wrappers/gpy_quadrature_wrappers.py
```diff
@@ -162,26 +164,57 @@ def dKdiag_dx(self, x: np.ndarray) -> np.ndarray:
...
+        quadrature_kernel_emukit = QuadratureRBFnoMeasure(rbf_kernel=standard_kernel_emukit,
+                                                          integral_bounds=integral_bounds,
+                                                          variable_names=integral_name)
+
+    # finite bounds: measure: uniform only not implemented
+    else:
+        if isinstance(measure, UniformMeasure):
+            raise ValueError('Q-kernel for uniform measure and finite bounds not implemented yet.')
+        else:
+            raise ValueError('Currently only None measure supported with these integral bounds.')
```
apaleyes: I must be missing something obvious here. How can an exception be avoided? All branches seem to lead to one

Also, this last message isn't clear. What is `None` measure? And it would be better to name things explicitly intead of referring to `these bounds`

emukit/quadrature/kernels/integration_measures.py
```diff
@@ -0,0 +1,147 @@
...
+        Computes the density at point x
+        :param x: points at which density is computes, shape (num_points, dim)
+        :return: the density at x, shape (num_points, )
+        """
+        # check if points are inside the box
+        bounds_lower = np.array([b[0] for b in self.bounds])
+        bounds_upper = np.array([b[1] for b in self.bounds])
+        inside_lower = 1 - (x < bounds_lower)
+        inside_upper = 1 - (x > bounds_upper)
+        inside_upper_lower = (inside_lower * inside_upper).sum(axis=1) == x.shape[1]
```
apaleyes: These are cool tricks, but mixing comparisons and operations in one line is hard to read. Example: "How to properly understand first line? What (x < bounds_lower) returns, is it an array of True/False or something else? 1 minus this array is what, an array or a number?" And so on. Can you rewrite this in a bit more straightforward-to-read way?

emukit/quadrature/kernels/integration_measures.py
```diff
@@ -0,0 +1,147 @@
...
+        Meaningful box-bounds around the measure. Outside this box, the measure should be virtually zero.
+
+        :return: box in which the measure lies. List of D tuples, where D is the dimensionality and the tuples contain
+        the lower and upper bounds of the box i.e., [(lb_1, ub_1), (lb_2, ub_2), ..., (lb_D, ub_D)]
+        """
+        return self.bounds
+
+
+class IsotropicGaussianMeasure(IntegrationMeasure):
+    """The isotropic Gaussian measure"""
```
apaleyes: The previous measure seems clear and simple, this imho requires a reference, maybe a link or a more detailed explanation of what is this object

emukit/quadrature/kernels/integration_measures.py
```diff
@@ -0,0 +1,147 @@
...
+        self.dim = mean.shape[0]
+
+    @property
+    def full_covariance_matrix(self):
+        return self.variance * np.eye(self.dim)
+
+    def compute_density(self, x: np.ndarray) -> np.ndarray:
+        """
+        Computes the density at point x
+        :param x: points at which density is computes, shape (num_points, dim)
```
apaleyes:
```suggestion
        :param x: points at which density is computed, shape (num_points, dim)
```

emukit/quadrature/kernels/integration_measures.py
```diff
@@ -0,0 +1,147 @@
...
+
+        :return: box in which the measure lies. List of D tuples, where D is the dimensionality and the tuples contain
+        the lower and upper bounds of the box i.e., [(lb_1, ub_1), (lb_2, ub_2), ..., (lb_D, ub_D)]
+        """
+        # Note: the factor 10 is pretty arbitrary here. Investigate if we can find a formula that adapts to
+        # higher dimensions where most of the volume is located in a sphere rather than around the mean.
+        factor = 10
+        lower = self.mean - factor * np.sqrt(self.variance)
+        upper = self.mean + factor * np.sqrt(self.variance)
+        return [(lb, up) for lb, up in zip(lower, upper)]
```
apaleyes:
```suggestion
        return list(zip(lower, upper))
```

emukit/quadrature/kernels/integration_measures.py
```diff
@@ -0,0 +1,147 @@
...
+        return np.exp(- np.sum(scaled_diff ** 2, axis=1)) / factor
+
+    def get_box(self) -> List[Tuple[float, float]]:
+        """
+        Meaningful box-bounds around the measure. Outside this box, the measure should be virtually zero.
+
+        :return: box in which the measure lies. List of D tuples, where D is the dimensionality and the tuples contain
+        the lower and upper bounds of the box i.e., [(lb_1, ub_1), (lb_2, ub_2), ..., (lb_D, ub_D)]
+        """
+        # Note: the factor 10 is pretty arbitrary here. Investigate if we can find a formula that adapts to
```
apaleyes: Just curious if that actually matters much what factor is. Do we just need to pick something large enough so that it is (almost) guaranteed there is no data outside?

emukit/quadrature/kernels/quadrature_kernels.py
```diff
@@ -20,18 +21,36 @@ class QuadratureKernel:
...
         :param kern: standard emukit kernel
         :param integral_bounds: defines the domain of the integral. List of D tuples, where D is the dimensionality
         of the integral and the tuples contain the lower and upper bounds of the integral
-        i.e., [(lb_1, ub_1), (lb_2, ub_2), ..., (lb_D, ub_D)]
-        :param integral_name: the (variable) name(s) of the integral
-        """
+        i.e., [(lb_1, ub_1), (lb_2, ub_2), ..., (lb_D, ub_D)]. None if bounds are infinite
+        :param measure: the integration measure. Note that the integral in the QuadratureKernel are specific to the
+        measure. None if no measure.
+        :param variable_names: the (variable) name(s) of the integral
```
apaleyes: What are they for?

emukit/quadrature/kernels/quadrature_kernels.py
```diff
@@ -20,18 +21,36 @@ class QuadratureKernel:
     An example of a specific QuadratureKernel and IStandardKernel pair is QuadratureRBF and IRBF.
     """
 
-    def __init__(self, kern: IStandardKernel, integral_bounds: List[Tuple[float, float]], integral_name: str='') \
+    def __init__(self, kern: IStandardKernel, integral_bounds: Union[None, List[Tuple[float, float]]],
+                 measure: Union[None, IntegrationMeasure], variable_names: str= '') \
```
apaleyes:
```suggestion
                 measure: Union[None, IntegrationMeasure], variable_names: str='') \
```

emukit/quadrature/kernels/quadrature_kernels.py
```diff
@@ -20,18 +21,36 @@ class QuadratureKernel:
...
+        """
+
+        # we define reasonable box bounds for the integral because e.g., the optimizer of the acquisition function
+        # requires finite bounds. The box is defined by the integration measure. See each integration measure for
+        # details. Note that this only affects methods that use this box, e.g. the acqusition optimizers. The integrals
+        # in this kernel will have infinite bounds still.
+        if (integral_bounds is None) and (measure is None):
+            raise ValueError('integral_bounds and measure are both None. At least one of them must be given.')
+        if integral_bounds is None:
+            reasonable_box_bounds = measure.get_box()
```
apaleyes: Did you mean to use this value to initialize `self.integral_bounds`?

tests/emukit/quadrature/test_rbf_quadrature_kernel.py
```diff
@@ -4,25 +4,87 @@
...
+    assert isinstance(emukit_qrbf.qKq(), float)
+
+    # gradient shapes
+    assert emukit_qrbf.dKq_dx(x1).shape == (M1, D)
+    assert emukit_qrbf.dqK_dx(x2).shape == (D, M2)
+
+
+def test_rbf_qkernel_no_measure_qK(qrbf_no_measure):
+    emukit_qrbf, _, x2, _, _, _ = qrbf_no_measure
+    # to check the integral, we check if it lies in some confidence interval.
```
apaleyes: Does the code that did it still exist? May be worth adding it to the repo, even if it isn't called, and refer to it here. In case this computation needs to be repeated for some reason

tests/emukit/quadrature/test_rbf_quadrature_kernel.py
```diff
@@ -4,25 +4,87 @@
...
+    # these intervals were computed as follows: the kernel emukit_qrbf.K was integrated in the first argument by
+    # simple random sampling with 1e6 samples. This was done 100 times. The intervals show mean\pm 3 std of the 100
+    # integrals obtained by sampling. There might be a very small chance the true integrals lies outside the specified
+    # intervals.
+    intervals = np.array([[3.047326312081091, 3.076162828036314],
+                          [5.114069393071916, 5.144207478013706],
+                          [0.9879234032609836, 1.0000191615928034],
+                          [0.07073863074148745, 0.07217298756057355]])
```
apaleyes: I think this needs to be a bit larger, even up to 2 or 3 numbers after comma. With that precision things like numpy versions may cause the test to fail

tests/emukit/quadrature/test_measures.py
```diff
@@ -0,0 +1,54 @@
...
+    assert measure.dim == D
+
+    res = measure.compute_density(x)
+    assert res.shape == (N, )
+
+
+def test_iso_gauss_measure_invalid_input():
+    variance = -2.
+
+    with pytest.raises(ValueError):
```
apaleyes: Glad to see tests for errors! Would be great to have more of these, as you definitely have more failure cases for invalid input in the code

emukit/examples/vanilla_bayesian_quadrature_with_rbf/vanilla_bq_loop_with_rbf.py
mmahsereci: done

emukit/examples/vanilla_bayesian_quadrature_with_rbf/vanilla_bq_loop_with_rbf.py
mmahsereci: done

emukit/model_wrappers/gpy_quadrature_wrappers.py

emukit/model_wrappers/gpy_quadrature_wrappers.py
mmahsereci: done. Also added an ARD check as we need ARD=False.

emukit/quadrature/kernels/integration_measures.py
mmahsereci: done

emukit/quadrature/kernels/integration_measures.py
mmahsereci: do we call it abstract class if we don't use ABC?

emukit/quadrature/kernels/integration_measures.py
mmahsereci: done

emukit/quadrature/kernels/integration_measures.py
mmahsereci: I had put this check in case the integration measure is defined on a very small space, e.g., a 10-D integral over the bounds [0, 1e-12]. Then we would multiply 1e-12 10 times which might potentially be zero. This is nothing we can prevent. It is just an ill-conditioned integration problem. But it is easier to debug if there is a check here. 
I deleted the "It might be numerical problems..." part as it does not add additional information.

emukit/quadrature/kernels/integration_measures.py
mmahsereci: Yes, it should be float, but my editor thinks it's an array and highlights it. What should I do?

emukit/quadrature/kernels/integration_measures.py
mmahsereci: I added an explanation. It should be clear now.

emukit/quadrature/kernels/integration_measures.py
mmahsereci: I am not sure what I should add. It is an isotropic Gaussian with an isotropic Gaussian density. That's a pretty well defined term. The description of the variance in the docstring says in addition that the variance is scalar.  I could add a formula if you think that would help.

emukit/quadrature/kernels/integration_measures.py
mmahsereci: done

emukit/quadrature/kernels/integration_measures.py
mmahsereci: done

emukit/quadrature/kernels/integration_measures.py
mmahsereci: It is the box that restricts the measure to a reasonable finite box, i,e., outside the box the density should be pretty much zero. 10 std is quite a large box for a Gaussian, so it is pretty conservative. I changed the comment to the one below.

"Note: the factor 10 is somewhat arbitrary but well motivated. If this method is used to get a box for data-collection, the box will be 2x 10 standard deviations wide in all directions, centered around the mean. Outside the box the density is virtually zero. "

emukit/quadrature/kernels/quadrature_kernels.py
mmahsereci: what do you mean with "they"? It is not clear from the comment.

emukit/quadrature/kernels/quadrature_kernels.py
mmahsereci: done

tests/emukit/quadrature/test_rbf_quadrature_kernel.py
mmahsereci: I still have it. Where would I put it though?

tests/emukit/quadrature/test_rbf_quadrature_kernel.py
mmahsereci: Not sure what you mean by "this" in the first sentence. If you mean the floats in the intervals I think already 4-5 digits are enough as the rest is statistical noise anyways. I should perhaps round them even.

tests/emukit/quadrature/test_measures.py
mmahsereci: I added more tests

emukit/model_wrappers/gpy_quadrature_wrappers.py
mmahsereci: I added the type hint and renamed the function to 'create_emukit_model_from_gpy_model'

emukit/quadrature/interfaces/base_gp.py
mmahsereci: This is related to the discussion we had about the BQ interface. We can discuss offline.

emukit/quadrature/kernels/integration_measures.py
mmahsereci: done

emukit/model_wrappers/gpy_quadrature_wrappers.py
mmahsereci: point 1: yes, the first error is a placeholder for the kernel which will come after this PR. I had not included it because the PR was already big. 

point 2: I replaced it with "Currently only measure=None is supported with these integral bounds."

emukit/examples/vanilla_bayesian_quadrature_with_rbf/vanilla_bq_loop_with_rbf.py
```diff
@@ -4,44 +4,50 @@
...
+from emukit.quadrature.kernels.integration_measures import IntegrationMeasure
+from emukit.model_wrappers.gpy_quadrature_wrappers import convert_gpy_model_to_emukit_model
 
 
-def create_vanilla_bq_loop_with_rbf_kernel(X: np.ndarray, Y: np.ndarray, integral_bounds: List,
-                                           rbf_lengthscale: float=1.0, rbf_variance: float=1.0) -> \
-        VanillaBayesianQuadratureLoop:
+def create_vanilla_bq_loop_with_rbf_kernel(X: np.ndarray, Y: np.ndarray,
+                                           integral_bounds: Union[None, List[Tuple[float, float]]],
+                                           measure: Union[None, IntegrationMeasure],
```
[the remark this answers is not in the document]
mmahsereci: Union[None, ...] is correct (see [here](https://docs.python.org/3/library/typing.html#typing.Union)). We kind of use it wrongly in emukit to denote and 'and', not and 'or'. 

I will replace it with Optional[IntegrationMeasure] (see [here](https://docs.python.org/3/library/typing.html#typing.Optional)) to highlight that it is different to how we usually use Union.

emukit/quadrature/kernels/quadrature_kernels.py
mmahsereci: no, the integral bounds are infinity here. However, the optimizer required finite bounds, so I put a reasonable box only for the optimizer. For now, reasonable boxes are the ones outside of which the integration measure is, or virtually is, zero. The integrals are still computed with infinite bounds.

emukit/quadrature/kernels/integration_measures.py
apaleyes: absolutely, abstract class as a concept is just a part of OOO and is agnostic to language/library

emukit/quadrature/kernels/integration_measures.py
apaleyes: ah, ok. if there is an easy way to convince the editor, would be good to. otherwise just keep like this, no big deal

emukit/quadrature/kernels/integration_measures.py
apaleyes: a formula would definitely help. imagine a person who never saw term before. is there a way to write a docstring so that they wouldn't need to go to wiki while reading this code?

emukit/quadrature/kernels/integration_measures.py
apaleyes: clear for me now, thanks

emukit/quadrature/kernels/quadrature_kernels.py
apaleyes: i see. if github showed only 1 line it would be clearer. i talking about "variable_names" parameter

emukit/quadrature/kernels/quadrature_kernels.py
apaleyes: oh, i see now, thanks

tests/emukit/quadrature/test_rbf_quadrature_kernel.py
apaleyes: next to this test would be fine

tests/emukit/quadrature/test_rbf_quadrature_kernel.py
apaleyes: yes, the floats. 2-3 digits after comma would suffice, i think

emukit/quadrature/kernels/integration_measures.py
mmahsereci: ok, done.

emukit/quadrature/kernels/quadrature_kernels.py
mmahsereci: It is a string that gives the user the option to describe the variables with words. For instance I used the BQ model in the SEIR application and forgot which dimension stands for which physical parameter, e.g., dim0->infection rate, etc. A string is just a minimal version to add some kind of description in case someone is as forgetful as me. It is not used for anything inside the code.

emukit/quadrature/kernels/integration_measures.py
mmahsereci: OK, I'll add a formula as well as this phrase: " An isotropic Gaussian is a Gaussian with scalar co-variance matrix." 

Just for context, the measure is part of the problem/task. So it is a quantity that is designed by, and most likely well understood by the user.

tests/emukit/quadrature/test_rbf_quadrature_kernel.py
mmahsereci: I added 2 files, one for the qkernels, ine for the warped GP integrals

tests/emukit/quadrature/test_rbf_quadrature_kernel.py
mmahsereci: I did not round because it is very tedious to decide where to round. The numbers are stochastic anyways, so rounding should not affect if the test fails.

PR #250, opened 2019-10-17
Added multi point expected improvement acquisition function

esiivola:
*Issue #, if available:*
---
*Description of changes:*
- A new interface for using the gradients of model mean and covariance matrix with respect to input location
- Support for the new interface to GPyModelWrapper
- Multi point expected improvement acquisition function
- Tests for multi point expected improvement acquisition function (test acquisition function with sampling and gradients with numerical difference)

All tests pass


emukit/bayesian_optimization/acquisitions/expected_improvement.py
```diff
@@ -88,3 +90,338 @@ def get_standard_normal_pdf_cdf(x: np.array, mean: np.array, standard_deviation:
     pdf = scipy.stats.norm.pdf(u)
     cdf = scipy.stats.norm.cdf(u)
     return u, pdf, cdf
+
+
+class QExpectedImprovement(ExpectedImprovement):
```
apaleyes: Is that an accepted name in the literature? I mean, it isn't super descriptive, but it's ok to keep if that's what people use

emukit/bayesian_optimization/acquisitions/expected_improvement.py
```diff
@@ -88,3 +90,338 @@ def get_standard_normal_pdf_cdf(x: np.array, mean: np.array, standard_deviation:
...
+    def __init__(self, model: Union[IModel, IDifferentiable, IJointlyDifferentiable], jitter: float = float(0),
+                 fast_compute: bool = False, eps: float = 1e-3) -> None:
```
apaleyes: no need for float(0), just 0.0 is fine. also, i believe there shouldn't be space around equals when defining default values.

emukit/model_wrappers/gpy_model_wrappers.py
```diff
@@ -24,12 +24,13 @@ def __init__(self, gpy_model: GPy.core.Model, n_restarts: int = 1):
         self.model = gpy_model
         self.n_restarts = n_restarts
 
-    def predict(self, X: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
+    def predict(self, X: np.ndarray, full_cov: bool = False) -> Tuple[np.ndarray, np.ndarray]:
```
apaleyes: That's bad design in GPy, because it changes the structure of the output. Can you instead define a separate method that would internally call `self.model.predict(X, full_cov=True)`?

requirements/requirements.txt
```diff
@@ -3,6 +3,6 @@ numpy>=1.14.5
 # This is unfortunate - we don't need matplotlib
 # but until GPy and GPyOpt get their dependencies straight
 # we need GPy's plotting extra to ensure smooth installation
-GPy[plotting]>=1.9.8
+GPy[plotting]>=1.9.9
```
apaleyes: Why is that?

emukit/bayesian_optimization/acquisitions/expected_improvement.py
```diff
@@ -88,3 +90,338 @@ def get_standard_normal_pdf_cdf(x: np.array, mean: np.array, standard_deviation:
...
+                        dSigk_pi = (Sigk_dx[l, j, :, :][:, ineq][ineq, :]
+                                    + Sigk_dx_ii[l, j] / (Sigk_ii**2) * (Sigk_mi @ Sigk_mi.T)
+                                    - (1.0 / Sigk_ii) * (SigtCross + SigtCross.T))
+                        B3[l, j] = (B3[l, j] + Sigk_ik * phi_ik * ((GPhi_ik.T @ dck_pi)
+                                    + 0.5 * np.sum(HPhi_ik * dSigk_pi)))
+                grad = grad + B1 + B2 + B3
+        return grad
+
+
+def covZk(Sigma: np.ndarray, k: int) -> np.ndarray:
```
apaleyes: I appreciate that original might have been using these shortenings, but we can do better. Let's replace this cryptic stuff with proper names. There is really no reason to say "decompo" instead of "decompose", right?

emukit/bayesian_optimization/acquisitions/expected_improvement.py
```diff
@@ -88,3 +90,338 @@ def get_standard_normal_pdf_cdf(x: np.array, mean: np.array, standard_deviation:
...
+                                 dSigma_dx: np.ndarray, y_minimum: float, pk: np.ndarray,
+                                 symmetric_term: np.ndarray, fast_compute: bool = False,
+                                 eps: float = float(1e-3)) -> np.ndarray:
+        """
+        Computes the gradient of multi point Expected Improvement. A helper function for the class.
+
+        :param mu: Prediction mean at locations where the evaluation is done.
+        :param Sigma: Prediction covariance at locations where the evaluation is done.
+        :param dmu_dx: Prediction mean gradient at locations where the evaluation is done.
+        :param Sigma: Prediction covariance gradient at locations where the evaluation is done.
```
apaleyes: that's dSigma_dx, right?

emukit/bayesian_optimization/acquisitions/expected_improvement.py
```diff
@@ -88,3 +90,338 @@ def get_standard_normal_pdf_cdf(x: np.array, mean: np.array, standard_deviation:
...
+        y_minimum = np.min(self.model.Y, axis=0)
+        qei, pk, symmetric_term = QExpectedImprovement.get_acquisition(mean, variance, y_minimum,
+                                                                       self.fast_compute, self.eps)
+
+        mean_dx, variance_dx = self.model.get_joint_prediction_gradients(x)
+        qei_grad = QExpectedImprovement.get_acquisition_gradient(mean, variance, mean_dx, variance_dx, y_minimum,
+                                                                 pk, symmetric_term, self.fast_compute, self.eps)
+        return -qei, -qei_grad
+
+    @staticmethod
```
apaleyes: What's the point of making it static and passing instance info like eps? Same goes for the next one

emukit/bayesian_optimization/acquisitions/expected_improvement.py
```diff
@@ -88,3 +90,338 @@ def get_standard_normal_pdf_cdf(x: np.array, mean: np.array, standard_deviation:
...
+    def get_acquisition(mu: np.ndarray, Sigma: np.ndarray, y_minimum: float,
```
apaleyes: Would it be too much to ask to break this method down a bit? It's a bit too large tbh. Same goes for the next one

emukit/bayesian_optimization/acquisitions/expected_improvement.py
apaleyes: Another point about these helper methods. Let's clearly indicate them as private with leading underscores - i don't think they can be reused for anything else

tests/emukit/bayesian_optimization/test_q_expected_improvement.py
```diff
@@ -0,0 +1,66 @@
+import numpy as np
```
apaleyes: That's great to have some tests. We also have generic file to test acquisition functions. Would you be able to add QEI there too?

https://github.com/amzn/emukit/blob/master/tests/emukit/test_acquisitions.py

emukit/bayesian_optimization/acquisitions/expected_improvement.py
esiivola: The spaces are there because ExpectedImprovement-class has them and I didn't want to edit the existing code just for style issues. Should I remove the spaces also from ExpectedImprovement?

emukit/bayesian_optimization/acquisitions/expected_improvement.py
mmahsereci: The paper refers to it at as "multipoint Expected Improvement", then later in the text they shorten in to q-EI for brevity. For us (as we don't need brevity) "MultipointExpectedImprovement" might be the better choice.

requirements/requirements.txt
apaleyes: If that is so - yes, please, that would be great to fix in EI file too. Thanks!

emukit/bayesian_optimization/acquisitions/expected_improvement.py
apaleyes: If that is so - yes, please, that would be great to fix in EI file too. Thanks!

emukit/bayesian_optimization/acquisitions/expected_improvement.py
apaleyes: thanks for chiming in, @mmahsereci , MultipointExpectedImprovement is certainly more expressive

tests/emukit/bayesian_optimization/test_q_expected_improvement.py
esiivola: Oh, I didn't notice this before as it wasn't in the bayesian_optimization-folder. All acquisition functions in the source code are in the bayesian_optimization subfolder. Would it make sense to move this test to the correct folder?

emukit/bayesian_optimization/acquisitions/expected_improvement.py
esiivola: These helper methods are not inside the class as they are not using any class attributes. In order to indicate that they are private (with underscore), I would have to move them inside the class. I kinda made them private by only importing the class in the __init__.py-file, so that these functions are not visible if not importing this file directly.

If I would move them inside the class, it would be natural to define them as @staticmethod but then I couldn't make them private (Or could I?). Is there an elegant way to define private static methods? To my naive understanding, the current way is the closest I can get to it.

I agree that no one probably needs these methods anywhere, but on the other hand they are generic, but rarely used tricks to condition a multivariate normal distribution.

What do you think?

emukit/bayesian_optimization/acquisitions/expected_improvement.py
esiivola: You are right, there is no point in doing so. Thanks!

emukit/bayesian_optimization/acquisitions/expected_improvement.py
esiivola: You are absolutely right. I'll fix this,

apaleyes: thanks for explanation. i think it's fine as it is now, no need for more changes!

tests/emukit/bayesian_optimization/test_q_expected_improvement.py
apaleyes: this test is there because you can use it with any acquisition function, not only those used with BO. let's keep it where it is

esiivola: After having a look, the test set "test:acquisitions" isn't suitable for testing acquisition functions that evaluate a single value for a batch of points. I'll keep the tests as they are.

emukit/bayesian_optimization/acquisitions/expected_improvement.py
esiivola: Fixed now

emukit/bayesian_optimization/acquisitions/expected_improvement.py
esiivola: Fixed now

emukit/model_wrappers/gpy_model_wrappers.py
esiivola: Fixed now

emukit/bayesian_optimization/acquisitions/expected_improvement.py
esiivola: Fixed now

emukit/bayesian_optimization/acquisitions/expected_improvement.py
esiivola: Fixed now

emukit/bayesian_optimization/acquisitions/expected_improvement.py
esiivola: Fixed now

tests/emukit/bayesian_optimization/test_multipoint_expected_improvement.py
```diff
@@ -0,0 +1,68 @@
...
+
+    x0 = np.array([0.45, 0.55])
+    _check_grad(MultipointExpectedImprovement(model), TOL_GRAD, x0)
+    _check_grad(MultipointExpectedImprovement(model, fast_compute=True, eps=1e-3), TOL_GRAD_FAST, x0)
+
+def _check_grad(lp, tol, x0):
+    grad_error = check_grad(lambda x: lp.evaluate(x[:, None]).flatten(),
+                            lambda x: lp.evaluate_with_gradients(x[:, None])[1].flatten(), x0)
+    print(grad_error)
+    #assert np.all(grad_error < tol)
```
apaleyes: doesn't look right

emukit/model_wrappers/gpy_model_wrappers.py
```diff
@@ -132,6 +153,56 @@ def fix_model_hyperparameters(self, sample_hyperparameters: np.ndarray) -> None:
...
+            dsigma[:, :, i, j] = dKss_dxi - Ks @ w_inv @ K.T - K @ w_inv @ Ks.T
+    return dsigma
+
+
+def dmean(x: np.ndarray, X: np.ndarray, kern: GPy.kern, w_vec: np.ndarray) -> np.ndarray:
+    """
+    Compute the derivative of the posterior mean with respect to prediction input
+
+    :param x: Prediction inputs of shape (q, d)
+    :param X: Training inputs of shape (n, d)
```
apaleyes: that is easy to misread. can you give these two clearly different names?

emukit/model_wrappers/gpy_model_wrappers.py
```diff
@@ -132,6 +153,56 @@ def fix_model_hyperparameters(self, sample_hyperparameters: np.ndarray) -> None:
         self.model._trigger_params_changed()
 
 
+def dSigma(x: np.ndarray, X: np.ndarray, kern: GPy.kern, w_inv: np.ndarray) -> np.ndarray:
+    """
+    Compute the derivative of the posterior covariance with respect to the prediction input
+
+    :param x: Prediction inputs of shape (q, d)
+    :param X: Training inputs of shape (n, d)
```
apaleyes: As with dmean, these two need clearly different names

emukit/bayesian_optimization/acquisitions/expected_improvement.py
```diff
@@ -88,3 +90,425 @@ def get_standard_normal_pdf_cdf(x: np.array, mean: np.array, standard_deviation:
...
+        res = - (x - mu) / Sigma * scipy.stats.norm.pdf(x, mu, np.sqrt(Sigma))
+    else:
+        res = np.zeros((q, q))
+    for i in range(q - 1):
+        for j in range(i + 1, q):
+            res[i, j] = decompose_mvn(x, mu, Sigma, [i, j])
+
+    res = res + res.T  # Hessian matrix is symmetric
+    # diagonal terms can be computed with the gradient of CDF and the other hessian terms
+    if (gradient is None):
```
apaleyes:
```suggestion
    if gradient is None:
```

emukit/bayesian_optimization/acquisitions/expected_improvement.py
```diff
@@ -88,3 +90,425 @@ def get_standard_normal_pdf_cdf(x: np.array, mean: np.array, standard_deviation:
...
+
+def Phi_hessian(x: np.ndarray, mu: np.ndarray, Sigma: np.ndarray, gradient: np.ndarray=None) -> np.ndarray:
+    """
+    Compute hessian matrix of CDF of multivariate Gaussian distribution.
+
+    :param x: Location where the Hessian is evaluated.
+    :param mu: Mean of the multivariate Gaussian.
+    :param Sigma: Covariance of the multivariate Gaussian.
+    :param gradient: Gradient of the multivariate Gaussian at x.
+    """
```
apaleyes: Missing return

emukit/bayesian_optimization/acquisitions/expected_improvement.py
```diff
@@ -88,3 +90,425 @@ def get_standard_normal_pdf_cdf(x: np.array, mean: np.array, standard_deviation:
...
+def decompose_mvn(x: np.ndarray, mu: np.ndarray, Sigma: np.ndarray, k: list) -> np.ndarray:
+    """
+    Decompose Multivariate normal to probability of some random variables being smaller than the rest and multiply this
+    with the marginal probability distribution of those random variables.
+
+    :param x: Location of decomposition.
+    :param mu: Mean of the multivariate Gaussian.
+    :param Sigma: Covariance of the multivariate gaussian.
+    :param k: Indices of decomposition.
+    """
```
apaleyes: Missing return

emukit/bayesian_optimization/acquisitions/expected_improvement.py
```diff
@@ -88,3 +90,425 @@ def get_standard_normal_pdf_cdf(x: np.array, mean: np.array, standard_deviation:
...
+
+def get_correlations_given_value_of_i(b: np.ndarray, mu: np.ndarray, Sigma: np.ndarray, i: int) -> np.ndarray:
+    """
+    Partial correlations of variables k knowing the value of variable i.
+
+    :param b: Known values.
+    :param mu: Mean of the multivariate Gaussian knowing i.
+    :param Sigma: Covariance of the multivariate gaussian knowing i.
+    :param i: known variable.
+    """
```
apaleyes: Missing return

emukit/bayesian_optimization/acquisitions/expected_improvement.py
```diff
@@ -88,3 +90,425 @@ def get_standard_normal_pdf_cdf(x: np.array, mean: np.array, standard_deviation:
...
+    return res
+
+
+def get_Sigma_given_value_of_i(Sigma: np.ndarray, i: int) -> np.ndarray:
+    """
+    Covariances of variables k knowing the value of variable i.
+
+    :param Sigma: Covariance of the multivariate Gaussian.
+    :param i: Known variable.
+    """
```
apaleyes: Missing return

emukit/bayesian_optimization/acquisitions/expected_improvement.py
```diff
@@ -88,3 +90,425 @@ def get_standard_normal_pdf_cdf(x: np.array, mean: np.array, standard_deviation:
...
+        res[k, i] = Sigma[k, k] - Sigma[k, i]
+        res[i, k] = res[k, i]
+    for i in neqi:
+        for j in [j for j in range(i, q) if j is not k]:
+            res[i, j] = Sigma[i, j] + Sigma[k, k] - Sigma[k, i] - Sigma[k, j]
+            res[j, i] = res[i, j]
+    return res
+
+
+def get_Sigma_given_value_of_i(Sigma: np.ndarray, i: int) -> np.ndarray:
```
apaleyes: Eh? get sigma given sigma? let's find a better name here

tests/emukit/bayesian_optimization/test_multipoint_expected_improvement.py
```diff
@@ -0,0 +1,68 @@
...
+
+# Tolerance needs to be quite high since the q-EI is also an approximation.
+TOL = 5e-3
+# Tolerance for the gradient can be smaller since the approximation is not stochastic
+TOL_GRAD = 1e-6
+# Tolerance for the gradient of the fast method needs to be higher as it is an approximation of an approximation
+TOL_GRAD_FAST = 1e-4
+
+
+def test_acquisition_qei():
```
apaleyes: qei here and below needs to be renamed

tests/emukit/bayesian_optimization/test_multipoint_expected_improvement.py
esiivola: sorry about this! I totally missed this when committing...

PR #265, opened 2019-11-14
Simple bayesian monte carlo

mmahsereci:
*Issue #, if available:*

*Description of changes:*

- adding Simple Bayesian Monte Carlo (SBMC) as in Rasmussen & Ghahramani 2003
- added sampling capability to probability measures as this is needed for SBMC
- added some tests
- fixed some relative imports & made integration test look nicer


emukit/quadrature/kernels/integration_measures.py
```diff
@@ -42,8 +44,28 @@ def get_box(self) -> List[Tuple[float, float]]:
         """
         raise NotImplementedError
 
+    @property
+    def can_sample(self):
+        """True if combined with IntegrationMeasureSampler"""
```
apaleyes: not sure what 'combined with IntegrationMeasureSampler' means here, can you elaborate?

emukit/quadrature/kernels/integration_measures.py
```diff
@@ -42,8 +44,28 @@ def get_box(self) -> List[Tuple[float, float]]:
...
+    """Augments the integration measure with sampling capabilities. This might be needed for some candidate point
+    collection schemes."""
+
+    def get_samples(self, num_samples: int, context_manager: ContextManager=None) -> np.ndarray:
+        """
+        samples from the probability distribution defined by the integration measure.
+        :param num_samples: number of samples
+        :param context_manager: The context manager Contains variables to fix. If a context is given, this method
+        samples from the conditional distribution.
+        :return: samples, shape (num_samples, dim)
```
apaleyes: what is "dim" here?

emukit/quadrature/kernels/integration_measures.py
```diff
@@ -105,8 +131,27 @@ def get_box(self) -> List[Tuple[float, float]]:
         """
         return self.bounds
 
+    def get_samples(self, num_samples: int, context_manager: ContextManager=None) -> np.ndarray:
+        """
+        samples from the uniform distribution.
+        :param num_samples: number of samples
+        :param context_manager: The context manager Contains variables to fix. If a context is given, this method
+        samples from the conditional distribution.
```
apaleyes: Is this comment actually true? It does not seem to line up with implementation below, at least to my eyes

emukit/quadrature/kernels/integration_measures.py
```diff
@@ -176,6 +225,23 @@ def get_box(self) -> List[Tuple[float, float]]:
         upper = self.mean + factor * np.sqrt(self.variance)
         return list(zip(lower, upper))
 
+    def get_samples(self, num_samples: int, context_manager: ContextManager=None) -> np.ndarray:
+        """
+        samples from the isotropic Gaussian distribution.
+        :param num_samples: number of samples
+        :param context_manager: The context manager Contains variables to fix. If a context is given, this method
+        samples from the conditional distribution.
```
apaleyes: ditto

emukit/quadrature/kernels/integration_measures.py
```diff
@@ -42,8 +44,28 @@ def get_box(self) -> List[Tuple[float, float]]:
...
-class UniformMeasure(IntegrationMeasure):
+class IntegrationMeasureSampler:
+    """Augments the integration measure with sampling capabilities. This might be needed for some candidate point
+    collection schemes."""
+
+    def get_samples(self, num_samples: int, context_manager: ContextManager=None) -> np.ndarray:
+        """
+        samples from the probability distribution defined by the integration measure.
+        :param num_samples: number of samples
+        :param context_manager: The context manager Contains variables to fix. If a context is given, this method
```
apaleyes: I feel like "The context manager Contains variables to fix." needs rephrasing. Something like "The context manager that contains variables to fix and the values to fix them to."

emukit/quadrature/kernels/integration_measures.py
```diff
@@ -105,8 +131,27 @@ def get_box(self) -> List[Tuple[float, float]]:
...
+        samples from the uniform distribution.
+        :param num_samples: number of samples
+        :param context_manager: The context manager Contains variables to fix. If a context is given, this method
+        samples from the conditional distribution.
+        :return: samples, shape (num_samples, dim)
+        """
+
+        D = len(self.bounds)
+        bounds = np.asarray(self.bounds)
+        samples = np.reshape(np.random.rand(num_samples * D), [num_samples, D]) * (bounds[:, 1] - bounds[:, 0]) \
```
apaleyes:
```suggestion
        samples = np.random.rand(num_samples, D) * (bounds[:, 1] - bounds[:, 0]) \
```

emukit/quadrature/kernels/integration_measures.py
```diff
@@ -176,6 +225,23 @@ def get_box(self) -> List[Tuple[float, float]]:
...
 
+    def get_samples(self, num_samples: int, context_manager: ContextManager=None) -> np.ndarray:
+        """
+        samples from the isotropic Gaussian distribution.
+        :param num_samples: number of samples
+        :param context_manager: The context manager Contains variables to fix. If a context is given, this method
+        samples from the conditional distribution.
+        :return: samples, shape (num_samples, dim)
+        """
+        samples = self.mean + np.sqrt(self.variance) * np.reshape(np.random.randn(num_samples * self.dim), [num_samples, self.dim])
```
apaleyes:
```suggestion
        samples = self.mean + np.sqrt(self.variance) * np.random.randn(num_samples, self.dim)
```

emukit/quadrature/loop/bayesian_monte_carlo_loop.py
```diff
@@ -0,0 +1,47 @@
...
+    used with the same interface as the active and adaptive learning schemes that depend explicitly or implicitly
+    (through hyperparameters) on the previous evaluations.
+    """
+    def __init__(self, model: WarpedBayesianQuadratureModel, model_updater: ModelUpdater=None):
+        """
+        :param model: a warped Bayesian quadrature method, e.g., VanillaBayesianQuadrature
+        :param model_updater: Defines how and when the quadrature model is updated if new data arrives. Defaults to
+        FixedIntervalUpdater. If the dummy updater NoopModelUpdater is used which does not update the model, the
+        collected nodes are stored in the loop state. When using NoopModelUpdater, you can update the model after the
+        loop ran by calling model.set_data(loop_state.X, loop_state.Y); model.optimize().
```
apaleyes: I feel like this part about NoopModelUpdater isn't very relevant to this particular loop. We can consider placing it somewhere in the docs, or maybe in the docstring of the named model updater, but in here it just left me scratching my head for some time trying to figure out what is it doing here

emukit/quadrature/loop/quadrature_point_calculators.py
```diff
@@ -0,0 +1,69 @@
...
+        """
+        :param model: The vanilla Bayesian quadrature model
+        """
+        self.parameter_space = parameter_space
+        self.model = model
+
+        # if measure is probability measure, check if it has sampling capabilities
+        if self.model.base_gp.kern.measure is not None:
+            if not self.model.base_gp.kern.measure.can_sample:
+                raise ValueError("The given probability measure has no method 'get_samples', but Simple Bayesian Monte "
```
apaleyes: I'd rephrase with something like "does not support sampling"

emukit/quadrature/loop/quadrature_point_calculators.py
```diff
@@ -0,0 +1,69 @@
...
+        :param context: Contains variables to fix through optimization of acquisition function. The dictionary key is
+                        the parameter name and the value is the value to fix the parameter to.
+        :return: (1 x n_dims) array of next inputs to evaluate the function at
+        """
+
+        # Lebesgue measure
+        if self.model.base_gp.kern.measure is None:
+            if context is None:
+                return self.parameter_space.sample_uniform(1)
+            else:
```
apaleyes: we should fix this somehow. of course not within this PR

emukit/quadrature/kernels/integration_measures.py
mmahsereci: IntegrationMeasureSampler is the class defined below. `can_sample` is sth like `has_gradients`, so there may or may not be sampling capabilities.

Now that I think of it I can also just add the method `get_samples`to IntegrationMeasure and not implement it if we cannot sample. Would that be better? That is how it is handled in the Acquisition class currently.

emukit/quadrature/kernels/integration_measures.py
mmahsereci: dimensionality

emukit/quadrature/kernels/integration_measures.py
mmahsereci: not sure I understand. I sample from the joint and then set the context. You can do this because the dimensions of the joint are independent. thus p(x, y) = p(x)p(y). But possibly I am not seeing sth..

emukit/quadrature/kernels/integration_measures.py
mmahsereci: see above

emukit/quadrature/kernels/integration_measures.py
mmahsereci: done

emukit/quadrature/kernels/integration_measures.py
mmahsereci: done

emukit/quadrature/loop/quadrature_point_calculators.py
mmahsereci: could you elaborate please? what should we fix? (I don't understand from the code snipet) Do you mean that we loop over dimensions which is kinda slow..

emukit/quadrature/kernels/integration_measures.py
mmahsereci: done

emukit/quadrature/loop/bayesian_monte_carlo_loop.py
mmahsereci: Well, first I had NoopModelUpdater because it safes a tremendous amount of compute. The model isn't being updated, nor optimized. Then I changed it to FixedIntervalUpdater in the end because it is kind of awkward to manually set the data and optimize the model after the loop ran (it's kind of even difficult to locate the data). Especially when benchmarking, run_loop should return the trained model.

So I thought in case someone collects many samples and wants to safe time they can use NoopModelUpdater here. A hint is nice since I don't think I would find that option otherwise. But I agree that the way I wrote it it might not convey that message... What do you think, shall I rephrase or you think such hints are not necessary?

emukit/quadrature/loop/quadrature_point_calculators.py
mmahsereci: done

mmahsereci:
with regards to general comment, I called it SimpleBayesianMonteCarlo because it is BayesianMonteCarlo with Simple Random Sampling as in Section 2.1. But I am open to naming suggestions.

apaleyes:
Let's just stick with BayesianMonteCarlo for now

emukit/quadrature/loop/quadrature_point_calculators.py
apaleyes: I meant handling of context is None

emukit/quadrature/loop/bayesian_monte_carlo_loop.py
apaleyes: Just mark this part of docsting clearly as a hint

emukit/quadrature/kernels/integration_measures.py
apaleyes: ok, let's add it to the doc

emukit/quadrature/kernels/integration_measures.py
apaleyes: it's fine to stick to approach similar to acqusition, with can_sample and get_samples

emukit/quadrature/kernels/integration_measures.py
mmahsereci: done

emukit/quadrature/kernels/integration_measures.py
mmahsereci: done

emukit/quadrature/loop/bayesian_monte_carlo_loop.py
mmahsereci: ok, done

PR #272, opened 2019-12-04
Additional stopping functionality

henrymoss:
I have added in an convergence based stopping criterion. We can now tell the loop to stop if two consecutive queries are closer (in euclidean norm) than a user-specified value.

I have also added functionality to use multiple stopping criterion, e.g.specifying  a convergence and max number of steps for the optimization. The loop will stop when any criterion is triggered.

This is demonstrated in Emukit-tutorial-basic-use-of-the-library.ipynb .

### DIFF
--- a/emukit/core/loop/outer_loop.py
+++ b/emukit/core/loop/outer_loop.py
@@ -54,18 +54,18 @@ class OuterLoop(object):
         self.loop_start_event = EventHandler()
         self.iteration_end_event = EventHandler()
 
-    def run_loop(self, user_function: Union[UserFunction, Callable], 
+    def run_loop(self, user_function: Union[UserFunction, Callable],
                  stopping_condition: Union[StoppingCondition, int, List[StoppingCondition]],
-                 context: dict=None) -> None:
+                 context: dict = None) -> None:
         """
         :param user_function: The function that we are emulating
-        :param stopping_condition: If integer - a number of iterations to run, if object (or list of such objects) - a stopping 
+        :param stopping_condition: If integer - a number of iterations to run, if object (or list of such objects) - a stopping
                         condition object that decides whether we should stop collecting more points
         :param context: The context is used to force certain parameters of the inputs to the function of interest to
                         have a given value. It is a dictionary whose keys are the parameter names to fix and the values
                         are the values to fix the parameters to.
         """
-        
+
         if not ((isinstance(stopping_condition, list) and
                 all([isinstance(condition, StoppingCondition) for condition in stopping_condition])) or
                 isinstance(stopping_condition, int) or isinstance(stopping_condition, StoppingCondition)):
@@ -79,7 +79,7 @@ class OuterLoop(object):
             stopping_condition = FixedIterationsStoppingCondition(stopping_condition + self.loop_state.iteration)
 
         if isinstance(stopping_condition, StoppingCondition):
-            stopping_condition = [stopping_condition]            
+            stopping_condition = [stopping_condition]
 
         _log.info("Starting outer loop")
 
--- a/emukit/core/loop/outer_loop.py
+++ b/emukit/core/loop/outer_loop.py
@@ -59,16 +59,20 @@ class OuterLoop(object):
                  context: dict = None) -> None:
         """
         :param user_function: The function that we are emulating
-        :param stopping_condition: If integer - a number of iterations to run, if object (or list of such objects) - a stopping
-                        condition object that decides whether we should stop collecting more points
+        :param stopping_condition: If integer - a number of iterations to run, or an object (or list of such objects) - a stopping
+                        condition object that decides whether we should stop collecting more points. When input is a list,
+                        we stop when any of the conditions are violated.
         :param context: The context is used to force certain parameters of the inputs to the function of interest to
                         have a given value. It is a dictionary whose keys are the parameter names to fix and the values
                         are the values to fix the parameters to.
         """
 
-        if not ((isinstance(stopping_condition, list) and
-                all([isinstance(condition, StoppingCondition) for condition in stopping_condition])) or
-                isinstance(stopping_condition, int) or isinstance(stopping_condition, StoppingCondition)):
+        is_int = isinstance(stopping_condition, int)
+        is_single_condition = isinstance(stopping_condition, StoppingCondition)
+        is_list_of_conditions = (isinstance(stopping_condition, list)
+                                 and all([isinstance(condition, StoppingCondition) for condition in stopping_condition]))
+
+        if not (is_int or is_single_condition or is_list_of_conditions):
             raise ValueError("Expected stopping_condition to be an int or a single/list of StoppingCondition instance(s),"
                              "but received {}".format(type(stopping_condition)))
 
--- a/emukit/core/loop/outer_loop.py
+++ b/emukit/core/loop/outer_loop.py
@@ -12,7 +12,7 @@ from .user_function_result import UserFunctionResult
 from .candidate_point_calculators import CandidatePointCalculator
 from .model_updaters import ModelUpdater
 from .user_function import UserFunction, UserFunctionWrapper
-from .stopping_conditions import StoppingCondition, FixedIterationsStoppingCondition
+from .stopping_conditions import StoppingCondition, FixedIterationsStoppingCondition, ConvergenceStoppingCondition
 
 
 import logging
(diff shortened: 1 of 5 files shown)


emukit/core/loop/outer_loop.py
```diff
@@ -54,18 +54,22 @@ def __init__(self, candidate_point_calculator: CandidatePointCalculator,
...
-                 context: dict=None) -> None:
+    def run_loop(self, user_function: Union[UserFunction, Callable],
+                 stopping_condition: Union[StoppingCondition, int, List[StoppingCondition]],
+                 context: dict = None) -> None:
         """
         :param user_function: The function that we are emulating
-        :param stopping_condition: If integer - a number of iterations to run, if object - a stopping condition object
-                                   that decides whether we should stop collecting more points
+        :param stopping_condition: If integer - a number of iterations to run, if object (or list of such objects) - a stopping
+                        condition object that decides whether we should stop collecting more points
```
apaleyes: This comment should explain how the list is handled (the fact that we use "or" and not say "and"). Don't be afraid to make it a huge multi-line thing

emukit/core/loop/outer_loop.py
```diff
@@ -54,18 +54,22 @@ def __init__(self, candidate_point_calculator: CandidatePointCalculator,
...
         :param context: The context is used to force certain parameters of the inputs to the function of interest to
                         have a given value. It is a dictionary whose keys are the parameter names to fix and the values
                         are the values to fix the parameters to.
         """
-        if not (isinstance(stopping_condition, int) or isinstance(stopping_condition, StoppingCondition)):
-            raise ValueError("Expected stopping_condition to be an int or a StoppingCondition instance, "
+
+        if not ((isinstance(stopping_condition, list) and
+                all([isinstance(condition, StoppingCondition) for condition in stopping_condition])) or
+                isinstance(stopping_condition, int) or isinstance(stopping_condition, StoppingCondition)):
```
apaleyes: This `if` is very hard to parse now! Can you break it down somehow? One idea is to create boolean variables for each condition. Something like that:
```
   is_int = isinstance(stopping_condition, int)
   is_single_condition = ...
   is_condition_list = ...
```
Don't have to be like that, the point is to make this `if` readable.

emukit/core/loop/stopping_conditions.py
```diff
@@ -22,7 +27,8 @@ class FixedIterationsStoppingCondition(StoppingCondition):
     """ Stops after a fixed number of iterations """
     def __init__(self, i_max: int) -> None:
         """
-        :param i_max: Maximum number of function observations within the loop, excluding initial points
+        :param i_max: Maximum number of function
+        observations within the loop, excluding initial points
```
apaleyes:
```suggestion
                observations within the loop, excluding initial points
```

normally you want some additional indent for multiline param docstrings like that

emukit/core/loop/stopping_conditions.py
```diff
@@ -31,4 +37,31 @@ def should_stop(self, loop_state: LoopState) -> bool:
         :param loop_state: Object that contains current state of the loop
         :return: True if maximum number of iterations has been reached
         """
-        return loop_state.iteration >= self.i_max
+        status = loop_state.iteration >= self.i_max
+        if status is True:
```
apaleyes: You can use a more informative name, like `should_stop`, and just do `if should_stop`, which reads nicer than `is status is True`

emukit/core/loop/stopping_conditions.py
```diff
@@ -31,4 +37,31 @@ def should_stop(self, loop_state: LoopState) -> bool:
...
+            _log.info("Stopped after {} Evaluations".format(self.i_max))
```
apaleyes:
```suggestion
            _log.info("Stopped after {} evaluations".format(self.i_max))
```

emukit/core/loop/stopping_conditions.py
```diff
@@ -31,4 +37,31 @@ def should_stop(self, loop_state: LoopState) -> bool:
...
+        return status
+
+
+class ConvergenceStoppingCondition(StoppingCondition):
+    """ Stops once we choose a point within eps of a previous
+    point (with respect to euclidean norm), a notion of convergence"""
```
apaleyes: Feels like this comment isn't finished

emukit/core/loop/stopping_conditions.py
```diff
@@ -31,4 +37,31 @@ def should_stop(self, loop_state: LoopState) -> bool:
...
+    def __init__(self, eps: float) -> None:
+        """
+        :param eps: minimum distance between
+        two consecutive x's to keep running the model
```
apaleyes: ditto

emukit/core/loop/stopping_conditions.py
```diff
@@ -31,4 +37,31 @@ def should_stop(self, loop_state: LoopState) -> bool:
...
+
+    def should_stop(self, loop_state: LoopState) -> bool:
+        """
+        :param loop_state: Object that contains current state of the loop
+        :return: True if consecutive x's are too close
+        """
+        if loop_state.iteration < 2:
+            # less than 2 evaluations so cannot calculate distance
+            return False
+        status = math.sqrt(sum((loop_state.X[-1, :] - loop_state.X[-2, :]) ** 2)) <= self.eps
```
apaleyes: anything wrong with `numpy.linalg.norm`? also, same point about boolean variable name

notebooks/Emukit-tutorial-basic-use-of-the-library.ipynb
```diff
@@ -66,9 +66,27 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 3,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "ename": "ImportError",
```
apaleyes: That's not nice! Can you rerun this notebook to get rid of errors (there is one more below)

tests/emukit/core/test_loop_steps.py
```diff
@@ -22,6 +25,50 @@ def test_fixed_iteration_stopping_condition():
...
+    loop_state_mock = mock.create_autospec(LoopState)
+    loop_state_mock.iteration = 3
+    loop_state_mock.X.return_value(np.array([[0], [1], [1.01]]))
+    assert(any([condition.should_stop(loop_state_mock)for condition in stopping_condition]) is True)
+
+    # check if we stop when we should due to criterion 2
+    loop_state_mock = mock.create_autospec(LoopState)
+    loop_state_mock.iteration = 6
+    loop_state_mock.X.return_value(np.array([[0], [10], [20], [30], [40], [50]]))
+    assert(any([condition.should_stop(loop_state_mock)for condition in stopping_condition]) is True)
```
apaleyes:
```suggestion
    assert(any([condition.should_stop(loop_state_mock) for condition in stopping_condition]) is True)
```

tests/emukit/core/test_loop_steps.py
```diff
@@ -22,6 +25,50 @@ def test_fixed_iteration_stopping_condition():
...
+    assert(stopping_condition.should_stop(loop_state_mock) is False)
+
+    # check if we stop when we should
+    loop_state_mock = mock.create_autospec(LoopState)
+    loop_state_mock.iteration = 5
+    loop_state_mock.X.return_value(np.array([[0], [1], [2], [3], [3.01]]))
+    assert(stopping_condition.should_stop(loop_state_mock) is True)
+
+
+def test_list_of_stopping_conditions():
```
apaleyes: for that test what should ideally be mocked is stopping condition itself. that is, you would have two mock conditions in a list, and manipulate their "should_stop" outputs to test various combinations (none fires, one fires, both fire)

emukit/core/loop/stopping_conditions.py
henrymoss: The problem with using np here, is that you end up with a np boolean object. This doesn't work with the assert statements in the tests.

I kept getting assert error False is False
 as one False was the numpy bool and one was the standard python bool

emukit/core/loop/outer_loop.py
mmahsereci: @apaleyes might it we worth doing a "OR" and "AND" stopping condition that handles multiple stopping conditions. So you could have multiple stopping conditions that go into the OrStoppingCondition for example. Then it would be easier to handle the inputs here.

apaleyes: @mmahsereci in the future maybe, and there might be some nice ways to implement that. but for now that seems overkill. let's deal with it once someone actually needs it

emukit/core/loop/stopping_conditions.py
apaleyes: not sure i am following. `linalg.norm` can certainly return a single float, and `float <= float` returns python boolean. what am i missing?

emukit/core/loop/outer_loop.py
apaleyes: after offline chat with @mmahsereci we've decided to cut an issue to do something more flexible here: #273 . @henrymoss please ignore it for now, it is certainly outside the scope of this PR

tests/emukit/core/test_loop_steps.py
apaleyes: btw, here is how you can make `should_stop` return different outputs in sequence: https://stackoverflow.com/questions/24897145/python-mock-multiple-return-values

emukit/core/loop/stopping_conditions.py
henrymoss: np.linalg.norm returns a numpy float. So the comparison provides a numpy bool

apaleyes: oh, ok, missed that, thanks. in that case how about just converting numpy float to python float?

henrymoss:
Made all the style changes.

henrymoss:
Thanks for that fix. I ran the notebook in a venv and it messed up the init. 

@apaleyes I have a question about using the batch (with local penalization) for a multi-fidelity optimization. It runs fine, but it is hard to diagnose if it actually works (i.e actually performing local penalization across all the fidelity space as well as the search space). Did you ever run any tests on using batch with multi-fidlelity models?

apaleyes:
@henrymoss i don't recall @marpulli or anyone else doing that, tbh

henrymoss:
Yeah so it works. But I'm not sure if it is just a fluke. I will investigate and get back to you!

apaleyes:
@henrymoss thanks! but when you do, let's open a new issue for it

Issue #278, opened 2020-01-15
Sobol experimental design


AnthonyLarroque:
Hello,

Since I think the sobol experimental design was not implemented in emukit but was in GPyOpt, I developed it for emukit following the LatinDesign model. Even if it is not a crucial contribution, I am sending it to you in .txt format if you would like to include it in the official version of emukit:

[sobol_design.txt](https://github.com/amzn/emukit/files/4064785/sobol_design.txt)

Kind regards,
Anthony Larroque

apaleyes:
Thanks! Any reason you didn't cut a PR for this?

AnthonyLarroque:
Hi @apaleyes,

First of all, because I am not fully comfortable with GitHub and the issue was the easiest way for me to do that. I know Git is useful but until now, I did not work a lot with that. I will have to learn it properly at some point. 

Secondly, because I also made some changes in the examples.gp_bayesian_optimization module to include other features in GPBayesianOptimization and if I am not wrong, these differences would also be included in the pull request. For clarity reasons, I just decided to send the sobol_design.

Regards,
Anthony Larroque

apaleyes:
Ok, thanks for your contribution, I will make sure to integrate it in emukit.

mmahsereci:
This has been merged . Thanks @AnthonyLarroque and @apaleyes. I am closing the issue.

Issue #219, opened 2019-06-03
Get rid of model_free designs


apaleyes:
While working on #216, I had an idea of simplifying the experimental_design package.

At the moment we have model_free and model_based parts in there. model_based is what emukit is all about, and i think this part is just fine. model_free currently has Random, which is a thin wrapper around `space.sample_uniform`, and Latin, which is a thin wrapper around [pyDOE](https://pythonhosted.org/pyDOE/randomized.html#latin-hypercube). Both are used essentially as just input generator utilities, and aren't used anywhere inside the library. We cannot even argue them to be a core part of emukit itself, as, like said above, they are just thin wrappers around something else.

I am going to argue that there is nothing they actually contribute to emukit. We already have random elsewhere, and Latin is a small bit of a well-design library pyDOE which our users may use if they so wish. We can even point them to it in the docs. So I suggest we remove these, and leave experimental_design package like that:

experimental_design
|_ interfaces
|_ acquisitions
|_ experimental_design_loop.py

essentially removing the whole model free/model based dichotomy from the library.

marpulli:
I'm not convinced that removing them is a good idea. They are certainly useful in our example notebooks for creating initial designs so I would expect them to be useful for customers too. 

Looking at the [code](https://github.com/amzn/emukit/blob/645eb88c07e4f3ec3fc6e7ff16e1428f4846720c/emukit/experimental_design/model_free/latin_design.py#L22) required to use pyDOE to create some samples, it is quite a few lines, and we do things like round the samples in case of discrete variables, which a customer could quite easily forget to do.

We could also consider renaming them so customers are more aware of their intended purpose - to create initial designs.

apaleyes:
Sounds reasonable.

Perhaps then another approach would be to further extend our recent addition of `ParameterSpace.sample_uniform` API? Something like that:

```
class ParameterSpace:
    def sample(self, n: int, design: InitialDesign.Random) -> np.ndarray:
    # other options include Latin, Sobol, maybe few more from pyDOE
```

thus moving the whole thing into `emukit.core.initial_designs`

mmahsereci:
my two cents: I agree with @marpulli that they are quite useful. I see them a bit like initial designs/sampling schemes in auto-diff frameworks (they usually have quite afew). It is not really what auto-diff is about but without them it would be much more of a hassle to use the library. 

Moving into core.initial_designs sounds reasonable, as the term (and the designs) are quite agnostic to the method they are being used with.

[...]

PR #280, opened 2020-01-16
Added sobol initial design

apaleyes:
*Issue #, if available:* #278

*Description of changes:* implementation of sobol initial design, provided by @AnthonyLarroque

### DIFF
--- a/emukit/core/initial_designs/__init__.py
+++ b/emukit/core/initial_designs/__init__.py
@@ -3,3 +3,4 @@
 
 from .latin_design import LatinDesign  # noqa: F401
 from .random_design import RandomDesign  # noqa: F401
+from .sobol_design import SobolDesign  # noqa: F401
--- a/emukit/core/initial_designs/sobol_design.py
+++ b/emukit/core/initial_designs/sobol_design.py
@@ -1,0 +1,40 @@
+import numpy as np
+try:
+    from sobol_seq import i4_sobol_generate
+except ImportError:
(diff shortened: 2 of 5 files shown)


mmahsereci (commented):
Hi Andrei, apologies for the delay. I have a general questions. Whenever we have optional dependencies such as sobol_sq here, I need to install it to import unrelated thing. E.g. if I run

```
from emukit.core.initial_designs import RandomDesign
```
which does not use SobolDesign, then I get the error
```
ImportError: sobol_seq needs to be installed in order to use sobol design
```

Additionally, if I am developer who really doesn't care about initial designs, then I need to install it sobol_seq etc to run all tests. I am not saying I have a better option, but I kind of don't want to accumulate optional dependencies which are not optional in the end. Any thoughts? Or sth I don't see? Thanks!

apaleyes:
Hey Maren. That's a very good catch, we should certainly fix that. Optional dependency should be optional for the end user. The only problem is, I've created this PR with a new branch, not from my fork, so cannot push anything else into it. Let's close this PR, and I'll recreate it from the fork, with this import issue fixed.

As for accumulating optional depedencies - personally i don't have a problem with that, if they offer great benefits for the users. we already has a discussion about initial designs #219 , and sobol sequeneces is a popular choice, so i think we should keep it in this case. but again, as you rightly said, optional should be optional.

mmahsereci:
hi Andrei, I agree we should keep the initial designs as they are very useful. I'll close the PR then. Thanks!

PR #283, opened 2020-02-27
Added sobol initial design

apaleyes:
*Issue #, if available:* #278 

*Description of changes:* Follow up on #280 . Fixed issues with optional dependencies for Latin and Sobol designs.


emukit/core/initial_designs/sobol_design.py
```diff
@@ -0,0 +1,38 @@
+import numpy as np
```
mmahsereci: licence header missing

emukit/core/initial_designs/sobol_design.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+        param parameter_space: The parameter space to generate design for.
+        """
+        super(SobolDesign, self).__init__(parameter_space)
+
+    def get_samples(self, point_count: int) -> np.ndarray:
+        """
+        Generates requested amount of points.
+        :param point_count: Number of points required.
+        :return: A numpy array with shape (point_count x space_dim)
```
mmahsereci: This does not describe what is in the array

emukit/core/initial_designs/sobol_design.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+        bounds = self.parameter_space.get_bounds()
+        lower_bound = np.asarray(bounds)[:, 0].reshape(1, len(bounds))
+        upper_bound = np.asarray(bounds)[:, 1].reshape(1, len(bounds))
+        diff = upper_bound - lower_bound
+
+        X_design = np.dot(i4_sobol_generate(len(bounds), point_count), np.diag(diff.flatten())) + lower_bound
```
mmahsereci: Somehow .flatten() always looks scary. Can you use diff[0, :]?

emukit/core/initial_designs/sobol_design.py
```diff
@@ -0,0 +1,38 @@
...
+
+        samples = self.parameter_space.round(X_design)
```
mmahsereci: Why is rounding required? How much does it change the distribution of the samples? Is this a common thing to do when producing sobol design points? Or do you round for numerical stability?

emukit/core/initial_designs/sobol_design.py
apaleyes: Thanks, added. Side note, the year should probably be updated across the lib.

emukit/core/initial_designs/sobol_design.py
apaleyes: Fixed there and in other model free design files

emukit/core/initial_designs/sobol_design.py
apaleyes: This won't round continuous parameters, they'll be left intact. But we need to make sure discrete and categorical hold valid values, and this is what rounding refers to here. Rules are down to each variable's implementation, but i believe they just look for a closest (euclidean norm) valid value

emukit/core/initial_designs/sobol_design.py
apaleyes: understandable. changed

emukit/core/initial_designs/sobol_design.py
mmahsereci: I see, thanks. Does this mean we might get the same value more than once for discrete variables? It might be worth adding a comment in case that causes problems to some models. What do you think?

apaleyes: I am not sure how to make this comment clear enough. Also, why is getting same point an issue?

mmahsereci: For example in GPs you might get ill-conditioned kernel matrices when points are identical. Debugging that may be a pain.
Would you be OK with adding sth like "Rounding may apply to discrete and categorical parameters."?

apaleyes: Sure, where would you like this added? Same rounding applies across all initial designs. Also, we have some comments in place on variables themselves: https://github.com/amzn/emukit/blob/master/emukit/core/discrete_parameter.py#L53

mmahsereci: Ah, perfect, sorry, didn't see that! Then nevermind.. no need for additional comment then.

apaleyes:
Thanks! You would it need to push this now though, I can't do it anymore

PR #286, opened 2020-03-31
added Latin and Sobol to initial design __init__

ekalosak:
2-line change to `core/initial_design/__init__.py`


apaleyes:
Thanks for the PR! We actually had this setup exactly that way before, but then realized that this results in some unwanted installation issues. Namely, both latin and sobol require optional dependecies installed, and if you run either of them, they will fail with a clear error. But the user is content with RandomDesign, beacuse of these `__init__` level imports, the errors will be triggered anyway, which is a poor user experience. Thus we decided to not include both these designs in `__init__`

ekalosak:
Got it - thanks for the explanation.

PR #287, opened 2020-04-01
fix: allow GPBayesianOptimization to set batch_size argument

sunnyszy:
* Description of changes: allow to change the batch_size argument in GPBayesianOptimization 


sunnyszy:
Hi,

I prefer this interface of external objective function evaluation. But I found it doesn't support `batch_size`. So what I have done is to modify the class and pass this information to the base class.

emukit/examples/gp_bayesian_optimization/single_objective_bayesian_optimization.py
```diff
@@ -89,7 +90,7 @@ def _acquisition_chooser(self):
 
     def suggest_new_locations(self):
         """ Returns one or a batch of locations without evaluating the objective """
-        return self.candidate_point_calculator.compute_next_points(self.loop_state)[0]
+        return self.candidate_point_calculator.compute_next_points(self.loop_state)
```
apaleyes, line 93: This change fails the tests. Why do you think it is needed?

sunnyszy: Hi Andrei,

Yes, I believe this change is needed. As the function name suggested (`compute_next_points`), the return can be multiple points. Therefore the return array should be of shape (n_points, n_dim).

I've fixed the test to check the return array shape.

apaleyes: Makes sense, thanks for clarifying!

emukit/examples/gp_bayesian_optimization/single_objective_bayesian_optimization.py
```diff
@@ -24,7 +24,7 @@ class GPBayesianOptimization(BayesianOptimizationLoop):
     def __init__(self, variables_list: list, X: np.array, Y: np.array, noiseless: bool = False,
                  acquisition_type: AcquisitionType = AcquisitionType.EI, normalize_Y: bool = True,
                  acquisition_optimizer_type: OptimizerType = OptimizerType.LBFGS,
-                 model_update_interval: int = int(1)) -> None:
+                 model_update_interval: int = int(1), batch_size: int = 1) -> None:
 
         """
         Generic class to run Bayesian optimization with GPyRegression model.
```
mmahsereci, line 30: can you add the new parameter to the docstring, please?

Issue #295, opened 2020-04-17
What is a reason to use scipy==1.1.0?


aeivazi:
Scipy latest version is `1.4.1` any specific reason to depend on such an old version of `scipy`?

apaleyes:
No, there is no specific reason. We can fix it, or if you wish to use this opportunity to submit a PR - feel free!

aeivazi:
Thanks @apaleyes for a quick response I will be happy to PR.
One more question: would you mind to relax requirements for `emcee` to low bound as well?

PR #300, opened 2020-04-20
Change mcmc_sampler to be compatible with emcee 2x and 3x

MashaNaslidnyk:
*Issue #, if available:* #295

*Description of changes:* See PR https://github.com/amzn/emukit/pull/299 which highlighted the issue. MCMC output format has changed in emcee>3; now it returns a State rather than a list. State is backwards compatible with a list. Hence the code change that works with both.

*To test with two emcee versions:*

```
# suppose your global emcee version is 2.x . This runs a test against emcee>3. Check with python3 -c "import emcee; print(emcee.__version__)"

conda create --name test_emcee_3  
conda install -n test_emcee_3 pytest pytest-cov
source activate test_emcee_3
conda install pip
pip install "emcee>=3" 
python3 -c "import emcee; print(emcee.__version__)"  #!! must be 3x
pip install GPy matplotlib mock pyDOE pytest-lazy-fixture sobol_seq
python3 -m pytest -x --cache-clear

source deactivate test_emcee_3  
```

### DIFF
--- a/emukit/samplers/mcmc_sampler.py
+++ b/emukit/samplers/mcmc_sampler.py
@@ -55,10 +55,14 @@ class AffineInvariantEnsembleSampler(McmcSampler):
         sampler = emcee.EnsembleSampler(n_samples, X_init.shape[1], log_p_function)
 
         # Burn-In
-        samples, samples_log, _ = sampler.run_mcmc(X_init, burn_in_steps)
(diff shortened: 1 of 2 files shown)


MashaNaslidnyk:
> That is a non-trivial bug to fix. Great job!

thanks for the review - I'll merge it now.

PR #304, opened 2020-05-25
bugfix: in the preferential batch BO example the acquisition function…

esiivola:
…s were previously minimized, not maximized

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/emukit/examples/preferential_batch_bayesian_optimization/pbbo/acquisitions/acquisition_function.py
+++ b/emukit/examples/preferential_batch_bayesian_optimization/pbbo/acquisitions/acquisition_function.py
@@ -65,7 +65,7 @@ class EmukitAcquisitionFunctionWrapper(Acquisition):
         :param x: points where the acquisition is evaluated.
         :return: acquisition function values
         """
-        return self.acquisitionFunction.evaluate(x, self.model)
+        return -self.acquisitionFunction.evaluate(x, self.model)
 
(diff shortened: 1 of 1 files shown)


mmahsereci:
Hi Eero, are you sure this is happening? GradientAcquisitionOptimizer flips the sign of the acquisition function  in its _optimize method.

esiivola:
Dear Maren. Thank you for taking a look! I am positive about the bugfix being the right thing to do. GradientAcquisitionOptimizer flips the sign and then minimizes the flipped function (= maximizes the original function). In the preferential batch Bayesian optimization example, all acquisition function evaluations were ALREADY returned as flipped versions (negative qEI) so we ended up minimizing the acquisition functions, which is not what we want.

Before building the wrapper for emukit, I was performing the optimization of the acquisition myself using "acq_fun_optimizer"-function (https://github.com/esiivola/emukit/blob/fe23f046b7e4b92d065bb71212a5e6f082432451/emukit/examples/preferential_batch_bayesian_optimization/pbbo/acquisitions/expectation_acquisition.py#L235), which minimized the "evaluate"-function (https://github.com/esiivola/emukit/blob/fe23f046b7e4b92d065bb71212a5e6f082432451/emukit/examples/preferential_batch_bayesian_optimization/pbbo/acquisitions/expectation_acquisition.py#L287).
 However, this "evaluate" function is maximized by emukit. The bugfix flips this function inside the emukit wrapper.

mmahsereci:
I see, thanks for the explanation!

PR #305, opened 2020-05-27
fixed context manager for multiple categorical variables in context

ClaireCp:
*Issue #, if available:*
In context_manager: when considering multiple categorical variables in context, appending (instead of extending) the encodings will lead to the context_values being a sequence of lists and floats. This becomes an issue when expanding the vector at 
x_expanded[:, np.array(self.context_idxs).astype(int)] = self.context_values
because we are trying to assign a sequence to a list

*Description of changes:*
changed append to extend in emukit.core.optimization.context_manager.py and added a corresponding test to show the bug

### DIFF
--- a/tests/emukit/core/optimization/test_context_manager.py
+++ b/tests/emukit/core/optimization/test_context_manager.py
@@ -43,9 +43,6 @@ def catg_context():
 def test_context_manager_catg_expand_vector(catg_space, catg_context):
     context_manager = ContextManager(catg_space, catg_context)
     x = np.array([[4., -1], [3, 0.]])
-    print(context_manager.non_context_idxs)
-    print(context_manager.context_idxs)
-
     x_expanded = context_manager.expand_vector(x)
     assert context_manager.space.dimensionality == 11
     assert x_expanded.shape == (2, context_manager.space.dimensionality)
--- a/tests/emukit/core/optimization/test_context_manager.py
+++ b/tests/emukit/core/optimization/test_context_manager.py
@@ -1,14 +1,13 @@
 import numpy as np
 import pytest
 
-from emukit.core import ContinuousParameter, ParameterSpace
+from emukit.core import ContinuousParameter, CategoricalParameter, ParameterSpace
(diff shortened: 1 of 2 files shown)


tests/emukit/core/optimization/test_context_manager.py
```diff
@@ -28,3 +27,27 @@ def test_context_manager_expand_vector(space, context):
...
+
+
+@pytest.fixture
+def catg_space():
+    return ParameterSpace([ContinuousParameter('x1', 0, 15), 
+                           CategoricalParameter('x2', OneHotEncoding([0, 1, 2, 3])),
+                           CategoricalParameter('x3', OneHotEncoding([1, 2, 3, 4, 5])),
+                           ContinuousParameter('x4', -2, 3)]) 
+
+@pytest.fixture
```
ekalosak, line 39: Are these utilities (catg_space, catg_context) defined as fixtures to indicate their reusability across tests? If not, would it not make more sense to initialize them inside test_context_manager_catg_expand_vector?

tests/emukit/core/optimization/test_context_manager.py
```diff
@@ -28,3 +27,27 @@ def test_context_manager_expand_vector(space, context):
...
+def catg_context():
+    return {'x2': 0, 'x3': 3}
+
+def test_context_manager_catg_expand_vector(catg_space, catg_context):
+    context_manager = ContextManager(catg_space, catg_context)
+    x = np.array([[4., -1], [3, 0.]])
+    print(context_manager.non_context_idxs)
+    print(context_manager.context_idxs)
```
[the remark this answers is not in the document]
ClaireCp: Yes absolutely, sorry about that!

tests/emukit/core/optimization/test_context_manager.py
ClaireCp: I made them fixtures in case more tests with multiple context variables need to be added, but can definitely change them to parameters, let me know!

PR #315, opened 2020-09-11
Bump scipy version

marpulli:
Just use latest scipy - it's causing conflicts with other projects to pin to such a specific version


ekalosak:
Wow that introduced some weird bugs. I'm curious what changed in scipy that causes these tests to fail :\ Any ideas?

mmahsereci:
I am not sure if this PR introduced the failure. The previous PR by Henry got the same error when calling `check_grad` (see `approx_derivative` in scipy.optimize._numdiff.py)

ekalosak:
Gotcha. Thank you for clarifying @mmahsereci.

apaleyes:
By looking at a few last builds, it looks as though at some point builds started pulling scipy 1.5 instead of 1.4, and started failing. Examples:

* This PR, failed: `Downloading scipy-1.5.2-cp37-cp37m-manylinux1_x86_64.whl (25.9 MB)`

* [Build](https://travis-ci.org/github/amzn/emukit/builds/694364181) 3 months ago, succeeded: `Downloading scipy-1.4.1-cp36-cp36m-macosx_10_6_intel.whl (28.5 MB)`

* [PR](https://github.com/amzn/emukit/pull/311) submitted by @henrymoss, failed: `Downloading scipy-1.5.1-cp36-cp36m-macosx_10_9_x86_64.whl (28.8 MB)`

apaleyes:
I have experimented locally and confirm that the error is caused by the scipy version. In the same virtual env, tests pass with `scipy==1.4.1` and fail with `scipy==1.5.1`.

apaleyes:
@marpulli perhaps you were modifying a different version of the code... right now what I see in changes is going from `scipy>=1.1.0` to `scipy=>1`. Which to me does not make much difference, and isn't very specific.

apaleyes:
Ping @marpulli. Are you happy with how reqs file looks right now, or shall we continue with this PR?

marpulli:
Hey, @apaleyes I was relaxing the constraint so that we would bump up to new minor versions automatically, but happy to set it to >=1.5 instead.

apaleyes:
Sorry Mark, I really don't follow. What's wrong with just >=1.1.0 as it is now? Esp since we seem to have addressed the issue that was causing the tests to fail, so >=1.5 is not really necessary.  Sorry for being slow on this one.

marpulli:
It's me being slow 😆 . I was thinking that `>=1.1.0` would allow any version `1.1.x`  but not `1.2.x` but that doesn't seem to be true. Closing!

apaleyes:
Phew! Thanks Mark! I actually had to do some research to convince myself this is the case. But seems like it is!

Issue #316, opened 2020-09-16
Tests fail with scipy >= 1.5


apaleyes:
At the moment tests pass with `scipy==1.4.1` and fail with `scipy==1.5.1`. That's weird.

Originally discussed in #315

ekalosak:
For convenience, here's the comparison for 1.5.1 v 1.4.1 in scipy: https://github.com/scipy/scipy/compare/v1.5.1...v1.4.1

mmahsereci:
closing, as this seems to be fixed with PR #318 .

PR #318, opened 2020-10-05
Fix dimentionality issues with scipy

apaleyes:
*Issue #, if available:* #316 

*Description of changes:* Turned out to be some inaccuracies with dimensions. Tested with both scipy 1.4.1 and 1.5.1, tests pass

### DIFF
--- a/tests/emukit/bayesian_optimization/test_local_penalization.py
+++ b/tests/emukit/bayesian_optimization/test_local_penalization.py
@@ -78,6 +78,6 @@ class MockModel(IModel):
 
 
 def _check_grad(lp, tol, x0):
-    grad_error = check_grad(lambda x: lp.evaluate_with_gradients(x[None, :])[0],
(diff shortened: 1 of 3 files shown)


mmahsereci:
Thanks @apaleyes , I am still a bit confused, how is this backwards compatible? It seems the test is accessing a different element of the return object of check_grad, so how does it still pass in 1.4.1 with the changes? Or are you saying there was always a bug in the test and 1.5.1 made it apparent? 
Also, the branch seems out of date. I can update it for you but perhaps you want to do it yourself so we avoid merge conflicts.

apaleyes:
Functions we were giving to the `check_grad` used to return 2x1 array: `[[1]]`. Now they return 1x1 array: `[1]`. Most likely scipy before could tolerate this slight dimensionality mismatch, but with 1.5 became more strict about it. That's my guess at what is going on, I haven't read the source code of both versions to compare. So to answer your question: yes, I think there was always a bug, but it got swallowed before, and now became apparent.

PR #320, opened 2020-10-19
made the preferential batch bayesian optimization example discoverable

esiivola:
*Issue #, if available:*

*Description of changes:*
 
When emukit is installed, the example "preferential_batch_bayesian_optimization" is now discoverable using emukit.examples.preferential_batch_bayesian_optimization. This allows the user to use its methods without extra work.

Also, fixed few syntax issues.

### DIFF
--- a/emukit/examples/preferential_batch_bayesian_optimization/__init__.py
+++ b/emukit/examples/preferential_batch_bayesian_optimization/__init__.py
@@ -1,1 +1,1 @@
-from .pbbo import *
+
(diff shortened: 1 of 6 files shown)


emukit/examples/preferential_batch_bayesian_optimization/__init__.py
```diff
@@ -0,0 +1 @@
+from .pbbo import *
```
apaleyes, line 1: I don't think this is needed. What is needed though is the __init__.py itself, but just empty

esiivola: You are right, fixed!

apaleyes (approved):

Issue #321, opened 2020-10-26
>2 level NonLinearMultiFidelityModel incorrect prediction on levels above 2nd


rns294:
Running a test with `NonLinearMultiFidelityModel` using 3 levels of code. The first two level predictions are accurate but the 3rd level does not interpolate the training points and has a large error/variance. I setup a 3 level model using [similar code from paraklas](https://github.com/paraklas/NARGP/blob/master/branin.py) that the emukit code was based on and was able to predict all 3 levels with high accuracy/low variance. 

> import numpy as np
> 
> from emukit.multi_fidelity.convert_lists_to_array  import convert_x_list_to_array, convert_xy_lists_to_arrays
> 
> from emukit.multi_fidelity.models.non_linear_multi_fidelity_model  import make_non_linear_kernels, NonLinearMultiFidelityModel
> 
> import GPy
> 
> from matplotlib import pyplot as plt
> 
> np.random.seed(12345)
> 
> 
> def fh(x):
>     return (6*x-2)**2*np.sin(12*x-4)
> 
> 
> def fm(x):
>     return 0.4*fh(x)-x-1
> 
> 
> def fl(x):
>     return 0.5*fh(x)+10*(x-0.5)-(5)
> 
> 
> Nts = 200
> x_plot = np.linspace(0, 1, Nts)[:, None]
> y_plot_l = fl(x_plot)
> y_plot_m = fm(x_plot)
> y_plot_h = fh(x_plot)
> 
> n0, n1, n2 = 6, 15, 40
> Xtrain = np.random.rand(100, 1)
> idx = np.random.permutation(100)
> 
> x_train_h = Xtrain[idx[0:n0], :]
> x_train_m = Xtrain[idx[0:n1], :]
> x_train_l = Xtrain[idx[0:n2], :]
> 
> y_train_l = fl(x_train_l)
> y_train_m = fm(x_train_m)
> y_train_h = fh(x_train_h)
> 
> X_train, Y_train = convert_xy_lists_to_arrays([x_train_l, x_train_m,  x_train_h], [y_train_l, y_train_m, y_train_h])
> 
> X_plot = convert_x_list_to_array([x_plot, x_plot, x_plot])
> X_plot_l = X_plot[:len(x_plot)]
> X_plot_m = X_plot[len(x_plot):-len(x_plot)]
> X_plot_h = X_plot[-len(x_plot):]
> 
> plt.figure(figsize=(12, 8))
[...]

PR #322, opened 2020-10-30
Update non_linear_multi_fidelity_model.py

rns294:
*Issue #321*

*Description of changes:*
Update prediction for previous fidelity during call to optimize (lines 328-329). Previous implementation only made a call to the predict method of the previous fidelity model. It did not properly propagate information from lowest level model to the current fidelity.

### DIFF
--- a/emukit/multi_fidelity/models/non_linear_multi_fidelity_model.py
+++ b/emukit/multi_fidelity/models/non_linear_multi_fidelity_model.py
@@ -325,7 +325,8 @@ class NonLinearMultiFidelityModel(IModel, IDifferentiable):
         # Optimize all models for all fidelities but lowest fidelity
         for i in range(1, self.n_fidelities):
             # Set new X values because previous model has changed
-            previous_mean, _ = self.models[i - 1].predict(self.models[i].X)
+            is_ith_fidelity = self.X[:, self._fidelity_idx] == i
+            previous_mean, _ = self._predict_deterministic(self.X[is_ith_fidelity, :-1], i)
             augmented_input = np.concatenate([self.models[i].X[:, :-1], previous_mean], axis=1)
             self.models[i].set_X(augmented_input)
 


apaleyes:
Thanks for the PR!

I can't immediately convince myself that this change is correct. Specifically, I am wandering if this code will work the same for as the old code for 2-fidelity case. Can you explain why this is the case?

rns294:
Certainly. 

In the previous implementation, once the first level was optimized, the predicted output of level 1 at level 2 would be used to optimize level 2. This was correct, but on the next iteration of the for loop for the level 3 model optimization (`i=2`), a call is only made to the predict method of level 2 at the points in level 3 (`self.models[i - 1].predict(self.models[i].X)`). This is inconsistent with the theory of Le Gratiet and Perdikaris. At level 3, there should be a prediction at the level 3 points (X_3) by the level 1 model, then that should be passed into the level 2 model for a prediction at the set (X_3, Y*_1(X_3)), and finally the prediction from level 2 is passed as the input to level 3. 

By modifying the code to make a call to `_predict_deterministic` this is inconsistency is fixed because the `_predict_deterministic` function calls the predict methods of all models up to the level of the current model. 

For the specific case you mentioned of a 2 level code, `i=1` so in the previous implementation:
`previous_mean, _ = self.models[0].predict(self.models[1].X)`
where as in the new implementation:
`is_ith_fidelity = self.X[:, self._fidelity_idx] == 1`
yields a bool array of the points in the 2nd fidelity model and:
`previous_mean, _ = self._predict_deterministic(self.X[is_ith_fidelity, :-1], 1)`
which makes a call to `_predict_deterministic` with `i=1`. The for loop in `_predict_deterministic` starts at 1 and thus does nothing. So a call to `_predict_deterministic` results in the following:
`mean, variance = self.models[0].predict(X)`
where is `X` is the points in the second fidelity model. This then returns `mean` and assigns it to `previous_mean` in `optimize`. This is identical to the previous implementation.

apaleyes (approved):

apaleyes:
Thanks for the explanation, that sounds correct and reasonable. Feel free to close the related issue if this PR resolves it.

PR #326, opened 2020-11-09
fixed gumbel sampler bug

henrymoss:
Fixed Gumbel sampler subroutine for MES and MUMBO.

We now fit the Gumbel (minima) distribution instead of the Gumbel (maxima) distribution, which is what we require for function minimisation.

I have also added some comments explaining where the equations come from to avoid future confusions on this issue.


apaleyes (approved):

apaleyes:
Looks good, thanks Henry. I also see that the build succeeded, but github lost it somehow: https://travis-ci.org/github/EmuKit/emukit/builds/742413881 . Will merge anyway

PR #327, opened 2020-11-12
Fixed bug in examples/fabolas/fmin.py.

NeoChaos12:
It seems that there was an artifact in the implementation of the fabolas optimizer in the given examples from a previous implementation of log-scale sampling of the dataset fraction. This is a small fix that removes the said artifact.


javiergonzalezh (approved):
Good catch, sounds good to me.

apaleyes (approved):
Going to trust you folks on this one. Also, just for the reference, github lost the build link, but the build passed: https://travis-ci.org/github/EmuKit/emukit/builds/743258043

Issue #329, opened 2020-11-14
Dynamic LCB


rvvincelli:
Hi!

I am looking for a way to bring in the regret-optimized version of the GP-UCB/LCB acquisition as described [here](https://arxiv.org/pdf/0912.3995.pdf) (th.1). I wonder if something like [this](https://gist.github.com/rvvincelli/6e72ab554f03c35e093fdfb1048be34f) is acceptable? The real question is what is the best way to keep track of the iterations, so the `t` variable; in this example we just override the `beta` property at each iteration, as `update_parameters` is the update hook called at the beginning of each iteration.

Please let me know! Thanks,

apaleyes:
We have LoopState object that already tracks iterations. Or if that isn't sufficient you can extend it with whatever additional data you need. E.g. for benchmarking purposes we [add a collection of metrics to it](https://github.com/EmuKit/emukit/blob/master/emukit/benchmarking/loop_benchmarking/benchmarker.py#L65). Would that work?

rvvincelli:
Hi @apaleyes , thanks for answering. Is that object accessible from inside the acquisition class? My point is more to get the information right into the acquisition, rather than getting it from the outside in the optimization loop. I am not sure if I get it, but could you perhaps just illustrate how you would implement the stateful acquisition as above? Thanks,

apaleyes:
oh, good point. we don't have immediate examples of it being used inside the acquisition. I imagine it is technically possible: you would need to implement new Point Calculator that calls your own acquisition and passes state in. Like another 10 lines of code.

Also, at the first glance i don't see anything wrong with your implementation of the Dynamic LCB

rvvincelli:
@apaleyes , thank you for the explanation. If you think it is a good idea I may open a PR, so that we get more feedback anyway. Let me know; thanks again!

apaleyes:
Up to you, really. Emukit will certainly benefit from having Dynamic UCB implementation. But PR in the core lib will require things like code readability and unit tests. If you are happy with that - give it a go for sure. If not, consider adding it into "examples", where requirements aren't as strict.

rvvincelli:
Hi @apaleyes , I'll think about it, thanks! I would say at least a comment, then if I understand how to write proper unit test a PR too. Tx!

rvvincelli:
Opted for an example, see [pr](https://github.com/EmuKit/emukit/pull/332). Thanks!

PR #332, opened 2020-11-23
Adds the Dynamic NLCB acquisition as an example

rvvincelli:
*Issue #329*

*Adds example of the dynamic variant of the negative LCB acquisition function as described in the original paper.*
cc @apaleyes 


emukit/examples/dynamic_negative_lower_confidence_bound/dnlcb.py
```diff
@@ -0,0 +1,28 @@
+from emukit.bayesian_optimization.acquisitions import NegativeLowerConfidenceBound
+from emukit.core.interfaces import IModel, IDifferentiable
+import numpy as np
+from typing import Union
+
+class DynamicNegativeLowerConfidenceBound(NegativeLowerConfidenceBound):
+    def __init__(self, model: Union[IModel, IDifferentiable], input_space_size: int, delta: float) -> None:
+        """
```
apaleyes, line 8: That's a very nice docstring here. Just one thing about it: please add reference to the paper.

emukit/examples/dynamic_negative_lower_confidence_bound/dnlcb.py
```diff
@@ -0,0 +1,28 @@
...
+        :param input_space_size: the size of the finite D grid on which the function is evaluated
+        :param delta: the exploration parameter determining the beta exploration coefficient; delta must be in (0, 1) and it is inversely related to beta
+        """
+        assert input_space_size > 0, "Invalid dimension provided"
+        assert 0 < delta < 1, "Delta must be in (0, 1)"
+        super().__init__(model)
+        self.input_space_size = input_space_size
+        self.delta = delta
+        self.iteration = 0
+    def optimal_beta_selection(self) -> float:
```
apaleyes, line 24: add some empty line(s) between methods for readability

emukit/examples/dynamic_negative_lower_confidence_bound/dnlcb.py
```diff
@@ -0,0 +1,28 @@
...
+        proportional to beta itself - the higher the delta the less explorative the selection will be.
+        Please consider that regret bounds based on the dynamic exploration coefficient only hold for selected kernel classes exhibiting boundedness and smoothness.
+        See the base class for references.
+        This class may also be taken as a reference for acquisition functions that dynamically update their parameters thanks to the update_parameters() hook.
+        :param model: The underlying model that provides the predictive mean and variance for the given test points
+        :param input_space_size: the size of the finite D grid on which the function is evaluated
+        :param delta: the exploration parameter determining the beta exploration coefficient; delta must be in (0, 1) and it is inversely related to beta
+        """
+        assert input_space_size > 0, "Invalid dimension provided"
+        assert 0 < delta < 1, "Delta must be in (0, 1)"
```
apaleyes, line 19: Within examples using asserts is probably fine, but in the library we normally raise exceptions. It's a better practice overall: they are easier to catch, can be inhereted and documented, and don't look like debugging in the code.

apaleyes (commented):
Two major comments:

1. Your iteration count is the second one, besides the count in the Loop State. Developers need to make sure the two don't go out of sync, which may or may not cause trouble. Besides, this one has an assumption of update_params method only being called once per iteration. 

What I was expecting for this implementation new PointCalculator that would create this specific acquisition and pass iteration number in when calling it. That would reuse the iteration count that already exists.

2. If you have any code that uses this acquisition, would be nice to have in here two, maybe as a notebook or just as a python file. Some people love to see examples in action.

emukit/examples/dynamic_negative_lower_confidence_bound/dnlcb.py
rvvincelli: Ok, I've left it as it is now since it's just an example, but you may return a `ValueException` if you like of course.

emukit/examples/dynamic_negative_lower_confidence_bound/dnlcb.py
rvvincelli: +1

emukit/examples/dynamic_negative_lower_confidence_bound/dnlcb.py
rvvincelli: The parent class is referenced, explicitly. This is found in the very same paper and I thought a repeated reference is notf fine.

rvvincelli:
> Thanks for the PR, glad you found time to take it beoynd the discussion in the issue.
> 
> Two major comments:
> 
> 1. Your iteration count is the second one, besides the count in the Loop State. Developers need to make sure the two don't go out of sync, which may or may not cause trouble. Besides, this one has an assumption of update_params method only being called once per iteration.
> 
> What I was expecting for this implementation new PointCalculator that would create this specific acquisition and pass iteration number in when calling it. That would reuse the iteration count that already exists.
> 
> 1. If you have any code that uses this acquisition, would be nice to have in here two, maybe as a notebook or just as a python file. Some people love to see examples in action.

Hi @apaleyes , thanks for your comments!

On the alternative construction with a point calculator, I do not get it unfortunately; but of course you may rewrite it if it not too much work (actually this is why I thought it's easier to start with an example). If the params update hook is invoked multiple times per iteration, no big deal, unless too many times (see docstring comment).

I have added a toy example; that's not very exciting, it is basically the test for some part of the original software where this acquisition is used.

Feel free to edit or amend any of the above of course. Thank you,

apaleyes (approved):

Issue #344, opened 2021-02-12
GPyModelWrapper.generate_huperparameters_samples changes the value of fixed parameters


BrunoKM:
Before starting collecting hyperparameter samples with HMC, the below code adds some random noise to the optimised MAP parameter array.

https://github.com/EmuKit/emukit/blob/f53b6cc7dea6602123f008c0adbb8cb0788ee760/emukit/model_wrappers/gpy_model_wrappers.py#L124-L138

Adding noise in this way, however, alters the value of the parameters which have been fixed with `constrain_fixed()`. For instance:
```
>>> # Specify GPy model with a constrained parameter
>>> kern = GPy.kern.RBF(1)
>>> kern['lengthscale'] = 0.01
>>> kern['lengthscale'].constrain_fixed()
>>> gpy_model = GPy.models.GPRegression(X, Y, kern)
>>> model = GPyModelWrapper(gpy_model)
>>> print(model.model['rbf.lengthscale'])
GP_regression.rbf.lengthscale:
Param([0.01])
>>> samples = model.generate_hyperparameters_samples()
>>> print(model.model['rbf.lengthscale'])
GP_regression.rbf.lengthscale:
Param([0.066441629009533])
```

The correct way to deal with this would be to add noise to the unconstrained parameters only in `generate_hyperparameters_samples()`:
```
# Add noise where parameters aren't fixed
noise = 1. + np.where(self.model._fixes_, np.random.randn(self.model.param_array.size) * 0.01, 0.0)
self.model.param_array[:] = self.model.param_array * noise
```

apaleyes:
Good catch! Not a fan of calling private fields like that, does GPy provide any public method to get the list of fixed params?

Either way, would you like to cut a PR?

BrunoKM:
I think there is some API to get and set unfixed parameters only. And yeah, I'd be happy to send through a small PR

PR #350, opened 2021-02-25
Fixes GPyModelWrapper.generate_hyperparameters_samples changes the value of fixed parameters

BrunoKM:
*Issue #, if available:* #344

*Description of changes:*

Fixes the referenced hyperparameter sampling issue by only adding random noise to the unfixed parameters before initiating HMC sampling.

### DIFF
--- a/emukit/model_wrappers/gpy_model_wrappers.py
+++ b/emukit/model_wrappers/gpy_model_wrappers.py
@@ -136,7 +136,8 @@ class GPyModelWrapper(IModel, IDifferentiable, IJointlyDifferentiable, ICalculat
         self.model.optimize(max_iters=self.n_restarts)
         unfixed_params = [param for param in self.model.flattened_parameters if not param.is_fixed]
         for param in unfixed_params:
-            param *= (1. + np.random.randn(param.size)) * 0.01
+            # Add jitter by multiplying with log-normal noise with mean 1 and standard deviation 0.01 
+            param *= np.random.lognormal(np.log(1. / np.sqrt(1.0001)), np.sqrt(np.log(1.0001)))
         hmc = GPy.inference.mcmc.HMC(self.model, stepsize=step_size)
         samples = hmc.sample(num_samples=n_burnin + n_samples * subsample_interval, hmc_iters=leapfrog_steps)
         hmc_samples = samples[n_burnin::subsample_interval]
--- a/emukit/model_wrappers/gpy_model_wrappers.py
+++ b/emukit/model_wrappers/gpy_model_wrappers.py
@@ -136,7 +136,7 @@ class GPyModelWrapper(IModel, IDifferentiable, IJointlyDifferentiable, ICalculat
         self.model.optimize(max_iters=self.n_restarts)
         unfixed_params = [param for param in self.model.flattened_parameters if not param.is_fixed]
         for param in unfixed_params:
-            param *= (1. + np.random.randn(param.size))
+            param *= (1. + np.random.randn(param.size)) * 0.01
         hmc = GPy.inference.mcmc.HMC(self.model, stepsize=step_size)
         samples = hmc.sample(num_samples=n_burnin + n_samples * subsample_interval, hmc_iters=leapfrog_steps)
         hmc_samples = samples[n_burnin::subsample_interval]
--- a/emukit/model_wrappers/gpy_model_wrappers.py
+++ b/emukit/model_wrappers/gpy_model_wrappers.py
@@ -134,9 +134,13 @@ class GPyModelWrapper(IModel, IDifferentiable, IJointlyDifferentiable, ICalculat
 
         """
         self.model.optimize(max_iters=self.n_restarts)
+        # Add jitter to all unfixed parameters. After optimizing the hyperparameters, the gradient of the
+        # posterior probability of the parameters wrt. the parameters will be close to 0.0, which is a poor
+        # initialization for HMC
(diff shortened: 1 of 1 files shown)


emukit/model_wrappers/gpy_model_wrappers.py
```diff
@@ -134,7 +134,10 @@ def generate_hyperparameters_samples(self, n_samples=20, n_burnin=100, subsample
 
         """
         self.model.optimize(max_iters=self.n_restarts)
-        self.model.param_array[:] = self.model.param_array * (1. + np.random.randn(self.model.param_array.size) * 0.01)
+        unfixed_params = [param for param in self.model.flattened_parameters if not param.is_fixed]
+        for param in unfixed_params:
+            # Add jitter by multiplying with log-normal noise with mean 1 and standard deviation 0.01 
+            param *= np.random.lognormal(np.log(1. / np.sqrt(1.0001)), np.sqrt(np.log(1.0001)))
```
apaleyes, line 140: Can this comment be expanded? Right now it just says what the code does, which isn't particuarly helpful. Instead it should ideally talk about why jitter is needed, and why it has to be done with log-normal noise instead of simple uniform.

The original code wasn't documented at all, but that was poor, so let's use the opporunity to improve here

apaleyes (commented):
Thanks! Just one ask on this

emukit/model_wrappers/gpy_model_wrappers.py
BrunoKM: To be perfectly honest, I am not entirely certain about why the jitter was needed in the first place.

I'm guessing that it's because the parameters are optimised first, which means that the gradient with respect to posterior probability will be close to `0`, and this would be a bad initialisation for HMC as it utilises the gradients.

I'm happy to add a comment saying this if my interpretation is correct.

BrunoKM: As for why I changed it to a log-normal, this is essentially to ensure the sign of the parameter value isn't flipped (in case it's constrained to be positive).

apaleyes: Both replies make sense to me, feel free to add them as comments

apaleyes: oh, you arealdy did. cool, thanks!

emukit/model_wrappers/gpy_model_wrappers.py
```diff
@@ -134,7 +134,14 @@ def generate_hyperparameters_samples(self, n_samples=20, n_burnin=100, subsample
...
         self.model.optimize(max_iters=self.n_restarts)
-        self.model.param_array[:] = self.model.param_array * (1. + np.random.randn(self.model.param_array.size) * 0.01)
+        # Add jitter to all unfixed parameters. After optimizing the hyperparameters, the gradient of the
+        # posterior probability of the parameters wrt. the parameters will be close to 0.0, which is a poor
+        # initialization for HMC
+        unfixed_params = [param for param in self.model.flattened_parameters if not param.is_fixed]
+        for param in unfixed_params:
+            # Add jitter by multiplying with log-normal noise with mean 1 and standard deviation 0.01 
+            # This ensures the sign of the parameter remains the same
+            param *= np.random.lognormal(np.log(1. / np.sqrt(1.0001)), np.sqrt(np.log(1.0001)), size=param.size)
```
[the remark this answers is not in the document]
BrunoKM: Yes, it does work because `param` is not a float or an array - it's an instance of a mutable type `GPy.core.parameterization.param.Param`:
```
In: type(model.flattened_parameters[0])
Out: GPy.core.parameterization.param.Param
```
Assigning an array/float to it changes it in place. Running the following snippet makes it clear:
```
In [3]: model = GPy.models.GPRegression(np.ones([3, 2]), np.ones([3, 1]), GPy.kern.Matern52(2, ARD=True))

In [4]: model["Mat52.variance"].fix()
Out[4]: array([0])

In [5]: model.param_array
Out[5]: array([1., 1., 1., 1.])

In [6]: # Add jitter to all unfixed parameters. After optimizing the hyperparameters, the gradient of the
   ...: # posterior probability of the parameters wrt. the parameters will be close to 0.0, which is a poor
   ...: # initialization for HMC
   ...: unfixed_params = [param for param in model.flattened_parameters if not param.is_fixed]
   ...: for param in unfixed_params:
   ...:     # Add jitter by multiplying with log-normal noise with mean 1 and standard deviation 0.01
   ...:     # This ensures the sign of the parameter remains the same
   ...:     param *= np.random.lognormal(np.log(1. / np.sqrt(1.0001)), np.sqrt(np.log(1.0001)), size=param.size)
   ...:

In [7]: model.param_array
Out[7]: array([1.        , 1.00995493, 1.01032573, 1.00097379])
```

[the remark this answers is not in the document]
BrunoKM: I personally also find this solution in GPy unintuitive, but after searching for quite a while, this the cleanest API call I found to get only the unfixed parameters without referencing private variables. Another way would be to call `model.parameter_names_flat(include_fixed=False)` to get the names of the viable parameters, but these names need to be sanitised first before they can be used to get the parameters as they're all prepended with `GPRegression.` (e.g. `GPRegression.Mat52.variance`)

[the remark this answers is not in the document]
apaleyes: confusing decision in GPy, i agree. but so be it, if that works

apaleyes (approved):

PR #358, opened 2021-03-10
adding sklearn gaussian process wrapper

charelstoncrabb:
Adding a model wrapper for scikit-learn's gaussian process regressor for a more object-oriented workflow


ekalosak:
Love the idea, looks like there's an undefined name in your wrapper though.
Also, I think the convention is to capitalize the first letter in python class names (https://www.python.org/dev/peps/pep-0008/#class-names)

charelstoncrabb:
> Love the idea, looks like there's an undefined name in your wrapper though.
> Also, I think the convention is to capitalize the first letter in python class names (https://www.python.org/dev/peps/pep-0008/#class-names)
> Finally, this needs a unit test or two - something like [this](https://github.com/EmuKit/emukit/blob/master/tests/emukit/examples/test_simple_gp_model.py) would probably be sufficient, but of course you could improve on the existing testing rigour and I'm sure nobody would complain :)

Sounds great! I'll put together those tests (and maybe think about some expanding!) and fix the imports/naming convention.

mmahsereci:
Hey, sorry for being late to the discussion. One thought I had: it might perhaps be better not to inherit from the sklearn class, but rather do sth like [here](https://github.com/EmuKit/emukit/blob/main/emukit/model_wrappers/gpy_model_wrappers.py#L21), and pass a sklearn instance to the wrapper. This would shield Emukit code more effectively, which was kind of the original idea of having wrappers, as currently sklearn specific model variables and methods are exposed. What do you think about this @apaleyes ?

apaleyes:
@charelstoncrabb I am picking up this PR to be able to merge it in, hope you don't mind!

PR #379, opened 2021-06-18
added square-root warping and factored out warpings in BQ

mmahsereci:
*Issue #, if available:*

*Description of changes:*
- this is another pre-PR for #375 
- factored out the warping from warped Gaussian process. We had discussed that previously @apaleyes  but it hadn't made sense back then as there was only 1 warping and you cannot mix and match warpings with models. However, now we need another warping that will be used by at least 2 methods, and you anyways cannot mix and match in BQ, so here we go..
- some docstring unifications
- added and `update_parameters` method to the warping model (will need that later).

### DIFF
--- a/emukit/quadrature/methods/warped_bq_model.py
+++ b/emukit/quadrature/methods/warped_bq_model.py
@@ -28,8 +28,8 @@ class WarpedBayesianQuadratureModel(IModel):
         """
         :param base_gp: The underlying Gaussian process model.
         :param warping: The warping of the underlying Gaussian process model.
-        :param X: The initial locations of integrand evaluations, shape (num_points, input_dim).
-        :param Y: The values of the integrand at X, shape (num_points, 1).
+        :param X: The initial locations of integrand evaluations, shape (n_points, input_dim).
+        :param Y: The values of the integrand at X, shape (n_points, 1).
         """
         self._warping = warping
         self.base_gp = base_gp
@@ -70,9 +70,9 @@ class WarpedBayesianQuadratureModel(IModel):
     def predict_base(self, X_pred: np.ndarray) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
         """Compute predictive means and variances of the warped GP as well as the base GP.
 
-        :param X_pred: Locations at which to predict, shape (num_points, input_dim).
+        :param X_pred: Locations at which to predict, shape (n_points, input_dim).
         :returns: Predictive mean and variances of warped GP, and predictive mean and variances of base-GP in that order
-                  all shapes (num_points, 1).
+                  all shapes (n_points, 1).
         """
         raise NotImplemented
 
@@ -80,9 +80,9 @@ class WarpedBayesianQuadratureModel(IModel):
                                                                              np.ndarray]:
         """Compute predictive means and covariance of the warped GP as well as the base GP.
 
-        :param X_pred: Locations at which to predict, shape (num_points, input_dim)
+        :param X_pred: Locations at which to predict, shape (n_points, input_dim)
         :returns: Predictive mean and covariance of warped GP, predictive mean and covariance of base-GP in that order.
-                  mean shapes both (num_points, 1) and covariance shapes both (num_points, num_points)
+                  mean shapes both (n_points, 1) and covariance shapes both (n_points, n_points)
         """
         raise NotImplemented
 
@@ -108,8 +108,8 @@ class WarpedBayesianQuadratureModel(IModel):
         First, the model parameters that are not being optimized are updated, as they may depend on the new data,
         then new data is set in the model.
 
-        :param X: Observation locations, shape (num_points, input_dim)
-        :param Y: Integrand observations at X, shape (num_points, 1)
+        :param X: Observation locations, shape (n_points, input_dim)
+        :param Y: Integrand observations at X, shape (n_points, 1)
         """
         self.update_parameters(X, Y)
         self.base_gp.set_data(X, self._warping.inverse_transform(Y))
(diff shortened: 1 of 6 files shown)


emukit/quadrature/methods/warped_bq_model.py
```diff
@@ -40,88 +42,93 @@ def X(self) -> np.ndarray:
...
+        self.base_gp.set_data(X, self._warping.inverse_transform(Y))
+
+    def update_parameters(self, X: np.ndarray, Y: np.ndarray) -> None:
+        """Update parameters of the model that are not being optimized. Use pass if no parameters need to be updated.
+
+        :param X: Observation locations, shape (num_points, input_dim)
+        :param Y: Integrand observations at X, shape (num_points, 1)
         """
-        self.base_gp.set_data(X, self.inverse_transform(Y))
+        raise NotImplementedError
```
apaleyes, line 123: Maybe you could just do `pass` here, and only those classes that need this override this method

emukit/quadrature/methods/warpings.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def update_parameters(self) -> None:
+        """No update for the identity transform."""
+        pass
+
+
+class SquareRootWarping(Warping):
+    """The square root warping"""
+
+    def __init__(self, offset: float, inverted: Optional[bool]=False):
```
apaleyes, line 55:
```suggestion
    def __init__(self, offset: float, is_inverted: Optional[bool]=False):
```

emukit/quadrature/methods/warpings.py
```diff
@@ -0,0 +1,91 @@
...
+
+        :param Y: function values of integrand, shape (num_points, 1).
+        :return: transformed values, shape (num_points, 1)
+        """
+        if self.inverted:
+            return np.sqrt(2. * (self.offset - Y))
+        else:
+            return np.sqrt(2. * (Y - self.offset))
+
+    def update_parameters(self, offset: Optional[float]=None) -> None:
```
apaleyes, line 85: Will this actually be used anywhere? It's very specific (talk about mix and match).

Perhaps it would make sense to do a different thing here? Make the warping object immutable, if you need to update, do something like

```
new_warping = SquaredRootWarping(new_offset, original_warping.inverted)
```

Just a wild suggestion here

emukit/quadrature/methods/warpings.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def inverse_transform(self, Y: np.ndarray) -> np.ndarray:
+        """Transform from integrand to base-GP.
+
+        :param Y: function values of integrand, shape (num_points, 1).
+        :return: transformed values, shape (num_points, 1)
+        """
+        raise NotImplemented
+
+    def update_parameters(self, **kargs) -> None:
```
apaleyes, line 24:
```suggestion
    def update_parameters(self, **kwargs) -> None:
```

tests/emukit/quadrature/test_warpings.py
```diff
@@ -0,0 +1,64 @@
+import numpy as np
```
apaleyes, line 1: Very nice testing suite!

apaleyes (commented):
Great PR, thanks Maren. Very nice clean up of the docstrings. But above all thank you for not jamming all those changes into a single huge PR, really appreciate it.

emukit/quadrature/methods/warped_bq_model.py
mmahsereci: OK, done.

emukit/quadrature/methods/warpings.py
mmahsereci: done

emukit/quadrature/methods/warpings.py
mmahsereci: done

emukit/quadrature/methods/warpings.py
mmahsereci: Yes, it;s needed by WSABI-L and WSABI-M. Both of these methods require the warping to be data dependent, so the warping changes after a new datapoint is collected (example [here](https://github.com/EmuKit/emukit/pull/375/files#diff-34e3d9470acc26afa739a9cf396ac75ca15dc290975d58ac165e97ed87620e53R63) in the draft-PR). 

So the options seem to be
1. Create a new immutable warping instance after each new datapoint is collected. This happens possibly a couple of thousand times. 
2. Create 1 instance of the warping and update its parameter after each new datapoint (currently this I do). 

What is better you think?

mmahsereci: Some more context: The current implementation is specific in the sense that `update_parameters` in the warping classes gets the new value and sets the new value of its own warping parameters (in this case `offset`). The default in the parent class I set to `pass`.  This methods is always called when the data is set in the parent class of all warped BQ models  [here](https://github.com/EmuKit/emukit/pull/379/files#diff-2fb1abe80576c9608cae8ae66a69543320695bbd669ce450e169c9b49ad53bcdR114). 

In case a specific BQ model (which inherits from said warped BQ model) requires a data dependent warping, the computation of the new value is done in the BQ model and not the warping. Example for WSABI-L method [here](https://github.com/EmuKit/emukit/pull/375/files#diff-34e3d9470acc26afa739a9cf396ac75ca15dc290975d58ac165e97ed87620e53R63). 
So warping is not convoluted with the specific BQ model. I opted for the `update_parameters` method instead of a setter for `offset` exactly for that reason, i.e., whenever a method requires the warping to be data-dependent, we do not need to change the warping code, *nor* `set_data` in the BQ method base class.

About mix and match: you cannot mix and match in BQ in the sense that to each BQ model there belongs a specific warping class, but the warping class need know know which BQ model it is being used with. Does that make sense?

mmahsereci: So, both options you suggest would work, but I guess it would not make the warping code less specific. I am wondering in general though: Is it better to create new instances frequently, or rather create one and update/set its parameters frequently (in case it has any)?

apaleyes: You're right, it's probably not worth the effort, these immutable instances don't help much.

apaleyes (approved):

Issue #359, opened 2021-03-13
Categorical Parameter as a context variable


mmilanek37:
I am trying to use a categorical variable as a context variable. When I compute_next_points with the candidate_point_calculator, I get the following error:

~\anaconda3\envs\intinker_ml\lib\site-packages\emukit\core\encodings.py in get_encoding(self, category)
     94             index = self.categories.index(category)
     95         except ValueError:
---> 96             raise ValueError("Unknown category {}" + str(category))
     97 
     98         return self.encodings[index].tolist()

ValueError: Unknown category {}[[1 0 0 0]]

However, when I check if the point is in the domain directly, it returns True:

categorical_variable.check_in_domain(np.array([[1, 0, 0, 0]]))

I can provide a sample Jupyter Notebook to help you understand the problem, but I am wondering if I am missing something here since an example doesn't exist for what I'm trying to do. I would also be willing to collaborate to make an example for creating a context variable with OneHotEncoding, as well as using the categorical parameter as a context variable.

apaleyes:
Yeah, categorical aren't easy because of back and forth conversion that has to be done. It can very well be inconsistency in our API. If you could share complete Colab notebook that reproduces the error, it would be very helpful

apaleyes:
Just pinging if you could share the codeto reproduce in this issue, @mmilanek37 . Thanks!

mmilanek37:
Thanks for the reminder--I made a workaround and have some code that I can share that would benefit the community. I'll have to take it out of the context of my project, but then I'll have that and sample code to reproduce the problem, probably near the end of the weekend

apaleyes:
I am going to close this, as the PR from @ndalchau that we just merged in enabled us to do `context = {'x': 'A'}`. If you feel there is still something lacking in the API, please do let us know either reopening this issue or creating a new one. Thanks!

Issue #319, opened 2020-10-18
categorical one hot encoding


untoreh:
calling `sample_uniform` on a onehot encoded category returns the encoded 1d array and not the category value, which breaks the space `sample_uniform`

apaleyes:
That doesn't sound very nice, thanks for reporting. Would you mind sharing the code you used to reproduce this issue?

untoreh:
```python
from emukit.core import CategoricalParameter, OneHotEncoding, OrdinalEncoding, ParameterSpace
from emukit.core.loop import RandomSampling, LoopState
cat_one = ["up", "down", "left", "right"]
cat_two = ["low", "medium", "high", "superhigh"]
space = [CategoricalParameter(name='one', encoding=OneHotEncoding(cat_one)),
         CategoricalParameter(name='two', encoding=OrdinalEncoding(cat_two)) ]
space = ParameterSpace(space)
loop = LoopState([])
points = RandomSampling(space)
points.compute_next_points(loop)
# array([[1., 0., 0., 0., 4.]])
```
The next points are passed to the `evaluate` method of the target function so I assume they should not be the encoded values, unless I am using it wrong? (since also ordinal values are not converted, so it's a problem with encodings in general)

btw, 
```python
# onehot with numpy
encodings = np.zeros(len(categories))
np.fill_diagonal(encodings, 1)

# ordinal..just np.arange(len(categories) (?)
```

apaleyes:
Yes, the output isn't decoded. That's intentional decision, because Emukit does not do modelling, and expects model as an input. We have no control over the way X is being put into the model. Therefore, to be as unopinionated as possible we decided to avoid doing encoding/decoding as a part of emukit's pipeline. That's essentially the trade-off between convenience and confusing behavior (which we could really fall into trying to cater for all possible use cases).

PR #381, opened 2021-06-28
Changes to categorical variable handling

sennendoko:

### DIFF
--- a/emukit/core/categorical_parameter.py
+++ b/emukit/core/categorical_parameter.py
@@ -53,24 +53,20 @@ class CategoricalParameter(Parameter):
     def dimension(self) -> int:
         return self.encodings.shape[1]
 
-    def check_in_domain(self, x: np.ndarray) -> bool:
+    def check_in_domain(self, x: str) -> bool:
         """
         Verifies that given values lie within the parameter's domain
 
-        :param x: 2d numpy array with shape (points, encoding) of points to check
+        :param x: a string representing the category
         :return: A boolean value which indicates whether all points lie in the domain
         """
-        if x.ndim != 2 or x.shape[1] != self.dimension:
-            raise ValueError("Expected x shape (points, {}), actual is {}"
-                             .format(self.dimension, x.shape))
+        if not isinstance(x, str):
+            raise ValueError("x should be a string representing the category")
 
-        for i, param in enumerate(self._cont_params):
-            # First check if this particular parameter is in domain
-            param_in_domain = param.check_in_domain(x[:, i])
-            if not param_in_domain:
-                return False
-
-        return True
+        if x in self.encoding.categories:
+            return True
+        else:
+            return False
 
     def sample_uniform(self, point_count: int) -> np.ndarray:
         """
--- a/emukit/core/encodings.py
+++ b/emukit/core/encodings.py
@@ -68,7 +68,7 @@ class Encoding(object):
         idx = (np.linalg.norm(self.encodings - x_row, axis=1)).argmin()
         return self.encodings[idx].copy()
 
-    def get_category(self, encoding: Union[List, np.ndarray]) -> str:
+    def get_category(self, encoding: Union[List, np.ndarray], raise_error = True) -> str:
         """
         Gets the category corresponding to the encoding.
 
@@ -78,7 +78,10 @@ class Encoding(object):
 
         indices = np.where(np.all(self.encodings == np.array(encoding), axis=1))
         if len(indices) == 0 or indices[0].size == 0:
-            raise ValueError("Given encoding {} does not correspond to any category".format(encoding))
+            if raise_error:
+                raise ValueError("Given encoding {} does not correspond to any category".format(encoding))
+            else:
+                return "Not identified category"
 
         category_idx = indices[0][0]
         return self.categories[category_idx]
@@ -99,7 +102,7 @@ class Encoding(object):
 
 
 class OneHotEncoding(Encoding):
-    def __init__(self, categories: List):
+    def __init__(self, categories: Union[List, np.ndarray]):
         """
         Initializes an instance of OneHotEncoding class
         and generates one hot encodings for given categories.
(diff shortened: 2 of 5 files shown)


apaleyes:
It's great to have this PR separated out, thanks again for doing that. I think it warrants a separate discussion most of all other changes.

The way categorical variables are handled in Emukit isn't very convenient, I have to agree with that. But that's the design choice we made, to avoid internal conversions that were very tricky and prone to bugs part of GPyOpt (just head to GPyOpt repo and lookup zip/unzip shenanigans). Since multiple encodings are possible, it is hard to keep proper track of categorical variable dimensionality when it isn't encoded. And that was with one certain model framework!

So in Emukit we offload handling of encoding to the user. Another reason is that Emukit takes model as an input, and in all likelihood the user has already dealt with encoding purely for modelling needs.

This is documented in [this notebook](https://github.com/EmuKit/emukit/blob/main/notebooks/Emukit-tutorial-select-neural-net-hyperparameters.ipynb) and [this issue](https://github.com/EmuKit/emukit/issues/319). A somewhat related discussion can also be found in [this PR](https://github.com/EmuKit/emukit/pull/191).

All of that considered, I might be overestimating the severity of these problems, so we can have a discussion on that here.

apaleyes:
So, given the previous comment, would you still like to pursue this change? If so, we would need to consider different design options and choose the most appripriate one. Also, there is a [notebook](https://github.com/EmuKit/emukit/blob/main/notebooks/Emukit-tutorial-select-neural-net-hyperparameters.ipynb) that may need updating if this change goes through.

Additionaly, would be interesting to hear the motivation behind this change.

ndalchau:
Hi @apaleyes , we've just learned that most of the changes in this PR are redundant. We were working in an old fork, and had added some changes to enable categorial parameters to be used in contexts. However, it seems that got resolved a year ago in  #305. With the other PRs integrated, we are now in a position to switch to the latest emukit. So, we can almost certainly close this PR.

apaleyes:
Perfect, thanks @ndalchau ! I am closing this PR, but being aware of the API's inconvenience I would be open to restart the conversation about these changes, if you feel the need again.

ndalchau:
Hi @apaleyes,

After looking at the code again, it appears some changes are still needed to get categorical parameters working properly within contexts.  

The main problem is as follows. Suppose you have a space containing one continuous and one categorical parameter:
```python
space = ParameterSpace([
    ContinuousParameter('x', 0, 1), 
    CategoricalParameter('y', OneHotEncoding(['A', 'B']))
])
```

Then eventually we choose to optimize `x` subject to `y` taking a value `A`. Nice, we can use a `Context`. e.g.
```python
context = {'y', 'A'}
```

I believe this is the correct way to initialize a context, because a `Context` is a `Dict[str, Any]`. 

However, when we pass this `context` to any `AcquisitionOptimiserBase`, the `optimize` method calls `_validate_context_parameters` (also takes `Dict[str, Any]`), which calls `check_in_domain` on all context values. When the context parameter is a categorical parameter, this `check_in_domain` method is expected to receive a `np.ndarray`. But what is being passed is a `str` (and in fact, can be `Any`).

I suspect `CategoricalParameter.check_in_domain` is expected to check the encoded parameter values, but it is being called on non-encoded parameters. 

My view is that this requires `check_in_domain` to work on `Any`, and to look up the values in the encoding (the approach of this PR). The only alternative I can think of is to change `_validate_context_parameters` to pass the encoded values. However, this would require those values existing in the domain, so would require a `check_in_domain` method. Hence why the former approach is required.

The problem can be identified by adding the following test (e.g. in `test_gradient_acquisition_optimizer.py`):
```python
def test_gradient_acquisition_optimizer_categorical_context(simple_square_acquisition):
    space = ParameterSpace([ContinuousParameter('x', 0, 1), CategoricalParameter('y', OneHotEncoding(['A', 'B']))])
    context = {'y': 'A'}
    optimizer = GradientAcquisitionOptimizer(space)
    opt_x, opt_val = optimizer.optimize(simple_square_acquisition, context)
```

Maybe we can add this test in the current PR, along with the other changes?

Please let us know what you'd prefer us to do to enable contexts to include categorical parameters, while still enabling use of `AcquisitionOptimizerBase`.

ndalchau:
Just noticed there are issues with categorical variables in contexts in issue #359.

apaleyes:
Thanks @ndalchau , now I understand what the issue is, and it also helped to clarify what was the matter in #359 . `check_in_domain` handling anything as an input sounds good, but might be a trifle complicated to get right. I'll have to give this a little think, but will come back with a better opinion.

ndalchau:
Thanks @apaleyes. I just thought of a simpler change.

Instead of assuming that the context_values are encoded values in `acquisition_optimizer.py`, we could do:
```python
if hasattr(param, 'encoding'):
    warn = context_value not in param.encoding.categories 
else: 
    warn = param.check_in_domain(context_value) is False
if warn:
    _log.warning(context_name + ' with value ' + str(context_value), ' is out of the domain')
else:
    _log.info('Parameter ' + context_name + ' fixed to ' + str(context_value))
```

apaleyes:
This proposed functionality is a good thing, it seems reasonable to allow users to use `context = {'y': 'A'}`. Therefore we need to:
  1. Check that the category we've been given is valid. If it isn't, just ignore it. Last fix from @ndalchau seems to be doing that.
  2. Make sure ContextManager can handle incorrect category. Test suit for ContextManager isn't very convincing right now.
  3. Consider merging validation into ContextManager. I am not entirely sure why they are separated, tbh. There might be a good reason though.

This seems most non-intrusive way, because it means we can leave categorical variable design as it is.

@ndalchau @sennendoko would any of you be interested in modifing this PR, or opening a new one, with this change?

Side note: `Dict[str, any]` as it is now in code seems a mistake. Here `any` is a Python function, which it probably should be `typing.Any`. Would be nice to incorporate this little fix.

PR #382, opened 2021-06-28
Changes to local penalization acquisition

sennendoko:
* Enabling Lipschitz constant and penalization ball reference point (f_min) to be fixed by the user.
* Enabling HMC samples to be used in local penalization (acquisition changes to IntegratedHyperParameterAcquisition). Previously, if using HMC, then only one of the HMC samples was used as a "representative point" in the evaluation of the acquisition, as opposed to integrating over those samples.
* Allowing control over optimization algorithm parameters for gradient_acquisition_optimizer.

### DIFF
--- a/emukit/bayesian_optimization/acquisitions/log_acquisition.py
+++ b/emukit/bayesian_optimization/acquisitions/log_acquisition.py
@@ -17,6 +17,10 @@ class LogAcquisition(Acquisition):
         """
         self.acquisition = acquisition
 
+    @property
+    def samples(self):
+        return self.acquisition.samples
+
     def evaluate(self, x: np.ndarray) -> np.ndarray:
         """
         :param x: Input location
--- a/emukit/bayesian_optimization/acquisitions/log_acquisition.py
+++ b/emukit/bayesian_optimization/acquisitions/log_acquisition.py
@@ -17,10 +17,6 @@ class LogAcquisition(Acquisition):
         """
         self.acquisition = acquisition
 
-    @property
-    def samples(self):
-        return self.acquisition.samples
-
     def evaluate(self, x: np.ndarray) -> np.ndarray:
         """
         :param x: Input location
--- a/emukit/bayesian_optimization/local_penalization_calculator.py
+++ b/emukit/bayesian_optimization/local_penalization_calculator.py
@@ -1,8 +1,9 @@
 import numpy as np
 import scipy.optimize
+from typing import Optional
 
 from ..core import ParameterSpace
-from ..core.acquisition import Acquisition
+from ..core.acquisition import Acquisition, IntegratedHyperParameterAcquisition
 from ..core.interfaces import IDifferentiable
 from ..core.loop import CandidatePointCalculator, LoopState
 from ..bayesian_optimization.acquisitions.local_penalization import LocalPenalization
@@ -21,7 +22,8 @@ class LocalPenalizationPointCalculator(CandidatePointCalculator):
     """
 
     def __init__(self, acquisition: Acquisition, acquisition_optimizer: AcquisitionOptimizerBase,
-                 model: IDifferentiable, parameter_space: ParameterSpace, batch_size: int):
+                 model: IDifferentiable, parameter_space: ParameterSpace, batch_size: int, 
+                 fixed_lipschitz_constant: Optional[float] = None, fixed_minimum: Optional[float] = None):
         """
         :param acquisition: Base acquisition function to use without any penalization applied, this acquisition should
                             output positive values only.
@@ -29,6 +31,8 @@ class LocalPenalizationPointCalculator(CandidatePointCalculator):
         :param model: Model object, used to compute the parameters of the local penalization
         :param parameter_space: Parameter space describing input domain
         :param batch_size: Number of points to collect in each batch
+        :param fixed_lipschitz_constant: User-specified Lipschitz constant, which controls influence of local penalization
+        :param fixed_minimum: User-specified minimum output, which specifies origin of penalization cones
         """
         if not isinstance(model, IDifferentiable):
             raise ValueError('Model must implement ' + str(IDifferentiable) +
@@ -39,8 +43,10 @@ class LocalPenalizationPointCalculator(CandidatePointCalculator):
         self.batch_size = batch_size
         self.model = model
         self.parameter_space = parameter_space
+        self.fixed_lipschitz_constant = fixed_lipschitz_constant
+        self.fixed_minimum = fixed_minimum
 
-    def compute_next_points(self, loop_state: LoopState, context: dict=None) -> np.ndarray:
+    def compute_next_points(self, loop_state: LoopState, context: dict = None) -> np.ndarray:
         """
         Computes a batch of points using local penalization.
 
@@ -51,21 +57,29 @@ class LocalPenalizationPointCalculator(CandidatePointCalculator):
         self.acquisition.update_parameters()
 
         # Initialize local penalization acquisition
-        local_penalization_acquisition = LocalPenalization(self.model)
+        local_penalization_acquisition = IntegratedHyperParameterAcquisition(self.model, LocalPenalization)
 
         # Everything done in log space so addition here is same as multiplying acquisition with local penalization
         # function.
         acquisition = self.acquisition + local_penalization_acquisition
 
         x_batch = []
-        for i in range(self.batch_size):
+        y_batch = []
+        for _ in range(self.batch_size):
             # Collect point
-            x_next, _ = self.acquisition_optimizer.optimize(acquisition, context)
+            x_next, y_next = self.acquisition_optimizer.optimize(acquisition, context)
             x_batch.append(x_next)
+            y_batch.append(y_next)
 
             # Update local penalization acquisition with x_next
-            f_min = np.min(self.model.Y)
-            lipschitz_constant = _estimate_lipschitz_constant(self.parameter_space, self.model)
+            if self.fixed_minimum is None:
+                f_min = np.min(np.append(self.model.Y, np.array(y_batch)))
+            else:
+                f_min = self.fixed_minimum
+            if self.fixed_lipschitz_constant is None:
+                lipschitz_constant = _estimate_lipschitz_constant(self.parameter_space, self.model)
+            else:
+                lipschitz_constant = self.fixed_lipschitz_constant
             local_penalization_acquisition.update_batches(np.concatenate(x_batch, axis=0), lipschitz_constant, f_min)
         return np.concatenate(x_batch, axis=0)
 
--- a/emukit/bayesian_optimization/local_penalization_calculator.py
+++ b/emukit/bayesian_optimization/local_penalization_calculator.py
@@ -1,4 +1,3 @@
-from emukit.core.interfaces.models import IPriorHyperparameters
 import numpy as np
 import scipy.optimize
 from typing import Optional
@@ -6,6 +5,7 @@ from typing import Optional
 from ..core import ParameterSpace
 from ..core.acquisition import Acquisition, IntegratedHyperParameterAcquisition
 from ..core.interfaces import IDifferentiable
+from ..core.interfaces.models import IPriorHyperparameters
 from ..core.loop import CandidatePointCalculator, LoopState
 from ..bayesian_optimization.acquisitions.local_penalization import LocalPenalization
 from ..core.optimization import AcquisitionOptimizerBase
--- a/emukit/bayesian_optimization/local_penalization_calculator.py
+++ b/emukit/bayesian_optimization/local_penalization_calculator.py
@@ -1,3 +1,4 @@
+from emukit.core.interfaces.models import IPriorHyperparameters
 import numpy as np
 import scipy.optimize
 from typing import Optional
@@ -57,7 +58,10 @@ class LocalPenalizationPointCalculator(CandidatePointCalculator):
         self.acquisition.update_parameters()
 
         # Initialize local penalization acquisition
-        local_penalization_acquisition = IntegratedHyperParameterAcquisition(self.model, LocalPenalization)
+        if isinstance(self.model, IPriorHyperparameters):
+            local_penalization_acquisition = IntegratedHyperParameterAcquisition(self.model, LocalPenalization)
+        else:
+            local_penalization_acquisition = LocalPenalization(self.model)
 
         # Everything done in log space so addition here is same as multiplying acquisition with local penalization
         # function.
--- a/emukit/bayesian_optimization/local_penalization_calculator.py
+++ b/emukit/bayesian_optimization/local_penalization_calculator.py
@@ -22,7 +22,7 @@ class LocalPenalizationPointCalculator(CandidatePointCalculator):
     """
 
     def __init__(self, acquisition: Acquisition, acquisition_optimizer: AcquisitionOptimizerBase,
-                 model: IDifferentiable, parameter_space: ParameterSpace, batch_size: int, 
+                 model: IDifferentiable, parameter_space: ParameterSpace, batch_size: int,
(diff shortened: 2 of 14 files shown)


emukit/core/optimization/gradient_acquisition_optimizer.py
```diff
@@ -52,10 +54,10 @@ def f_df(x):
             f_df = None
 
         optimizer = self._get_optimizer(context_manager)
-        anchor_points_generator = ObjectiveAnchorPointsGenerator(self.space, acquisition)
+        anchor_points_generator = ObjectiveAnchorPointsGenerator(self.space, acquisition, num_samples=self.num_samples)
 
         # Select the anchor points (with context)
-        anchor_points = anchor_points_generator.get(num_anchor=1, context_manager=context_manager)
+        anchor_points = anchor_points_generator.get(num_anchor=self.num_anchor, context_manager=context_manager)
```
apaleyes, line 60: That seems to change the default/previous behavior. Can `num_anchor` be 1 by default? Actually I suspect this change is not that severe to make a massive difference. What are your thoughts?

emukit/model_wrappers/gpy_model_wrappers.py
```diff
@@ -206,10 +205,10 @@ def dmean(x_predict: np.ndarray, x_train: np.ndarray, kern: GPy.kern, w_vec: np.
     """
     Compute the derivative of the posterior mean with respect to prediction input
 
-    :param x: Prediction inputs of shape (q, d)
-    :param X: Training inputs of shape (n, d)
+    :param x_predict: Prediction inputs of shape (q, d)
+    :param x_train: Training inputs of shape (n, d)
     :param kern: Covariance of the GP model
-    :param w_inv: Woodbury vector of the posterior fit of the GP
+    :param w_vec: Woodbury vector of the posterior fit of the GP
```
apaleyes, line 211: nice catch!

tests/emukit/multi_fidelity/test_non_linear_models.py
```diff
@@ -171,7 +171,7 @@ def wrap_func(x):
 
         def wrap_gradients(x):
             x_full = np.concatenate([x[None, :], [[2]]], axis=1)
-            return non_linear_model.get_prediction_gradients(x_full)[0]
+            return non_linear_model.get_prediction_gradients(x_full)[0].flatten()
```
apaleyes, line 174: Are these (this one and below) calls to `flatten()` needed? We are actually trying to avoid things like flatten, ravel and atleast_2d to keep good track of dimensions. Otherwise these dimensionality bugs are pretty hard to spot. Calls to these functions are still ok, but only for a good reason. What is the reason here?

emukit/core/acquisition/integrated_acquisition.py
```diff
@@ -29,12 +29,15 @@ def __init__(self, model: Union[IModel, IPriorHyperparameters], acquisition_gene
...
-                                                                   subsample_interval, step_size, leapfrog_steps)
+        self.update_parameters()
 
         acquisition = self.acquisition_generator(model)
         self._has_gradients = acquisition.has_gradients
 
+    @property
+    def samples(self):
+        return self.model.samples
+
```
apaleyes, line 40: This acquisition seems to be the only place that actually makes use of samples as a part of its state.

emukit/core/interfaces/models.py
```diff
@@ -74,16 +74,15 @@ def get_joint_prediction_gradients(self, X: np.ndarray) -> Tuple[np.ndarray, np.
 
 class IPriorHyperparameters:
     def generate_hyperparameters_samples(self, n_samples: int, n_burnin: int,
-                                         subsample_interval: int, step_size: float, leapfrog_steps: int) -> np.ndarray:
+                                         subsample_interval: int, step_size: float, leapfrog_steps: int) -> None:
         """
-        Generates the samples from the hyper-parameters of the model.
+        Generates the samples from the hyper-parameters of the model, and sets self.samples to that
```
apaleyes, line 79: I quite like the stateless nature of this method, I think it makes things explicit. As commented above, IntegratedAcquisition seem to be the only place that actually needs to keep samples as its state.

emukit/bayesian_optimization/local_penalization_calculator.py
```diff
@@ -51,21 +57,29 @@ def compute_next_points(self, loop_state: LoopState, context: dict=None) -> np.n
         self.acquisition.update_parameters()
 
         # Initialize local penalization acquisition
-        local_penalization_acquisition = LocalPenalization(self.model)
+        local_penalization_acquisition = IntegratedHyperParameterAcquisition(self.model, LocalPenalization)
```
apaleyes, line 60: That's a behavior change. How impactful is it? Is it worth thinking about preserving old behavior by default, and adding integration over hyperparametrs as an option? I don't have a strong opinion at the moment, just want to hear some thoughts. Also casting in @javiergonzalezh for discussion

javiergonzalezh: That makes sense. I think that the ideal here is to automatically detect if the model has hyper-priors and to enable the new behaviour (Integrated.. etc) when that is the case.

emukit/core/optimization/gradient_acquisition_optimizer.py
javiergonzalezh: That's a good point. The issue here is that using one anchor point is insufficient to optimize the acquisition properly, so it makes sense to be explicit about it and ask the user to select it (or use a larger number by default perhaps?)

emukit/model_wrappers/gpy_model_wrappers.py
```diff
@@ -330,24 +331,22 @@ def get_covariance_between_points(self, X1: np.ndarray, X2: np.ndarray) -> np.nd
         """
         return self.gpy_model.posterior_covariance_between_points(X1, X2, include_likelihood=False)
 
-    def generate_hyperparameters_samples(self, n_samples = 10, n_burnin = 5, subsample_interval  = 1,
+    def generate_hyperparameters_samples(self, n_samples = 10, n_burnin = 5, subsample_interval = 1,
                                          step_size = 1e-1, leapfrog_steps = 1) -> None:
         """
-        Generates the samples from the hyper-parameters
+        Generates the samples from the hyper-parameters, and sets self.samples to that (a numpy array whose rows are
+        samples of the hyper-parameters).
```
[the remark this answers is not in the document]
javiergonzalezh: I guess that the only reason to do that is if you want to cash the entire model after each iteration and reproduce the results. If not, I agree that saving the samples is not strictly needed.

tests/emukit/multi_fidelity/test_non_linear_models.py
javiergonzalezh: that's a good point, we need to do that.

emukit/core/optimization/gradient_acquisition_optimizer.py
ndalchau: I agree, it's almost always better to have `num_anchors` > 1. Our wrapper has its own control for this anyway, so it's not important for us, but perhaps we should change the `GradientAcquisitionOptimizer.init()` to have `num_anchors=1` by default instead of `num_anchors=5`, for consistency with the existing behaviour?

emukit/core/interfaces/models.py
ndalchau: We'll investigate storing the samples in `IntegratedHyperParameterAcquisition` instead of the model.

emukit/bayesian_optimization/local_penalization_calculator.py
```diff
@@ -39,8 +43,10 @@ def __init__(self, acquisition: Acquisition, acquisition_optimizer: AcquisitionO
         self.batch_size = batch_size
         self.model = model
         self.parameter_space = parameter_space
+        self.fixed_lipschitz_constant = fixed_lipschitz_constant
+        self.fixed_minimum = fixed_minimum
 
-    def compute_next_points(self, loop_state: LoopState, context: dict=None) -> np.ndarray:
+    def compute_next_points(self, loop_state: LoopState, context: dict = None) -> np.ndarray:
```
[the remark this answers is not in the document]
sennendoko: Done.

emukit/bayesian_optimization/local_penalization_calculator.py
```diff
@@ -51,21 +57,29 @@ def compute_next_points(self, loop_state: LoopState, context: dict=None) -> np.n
...
+            y_batch.append(y_next)
 
             # Update local penalization acquisition with x_next
-            f_min = np.min(self.model.Y)
-            lipschitz_constant = _estimate_lipschitz_constant(self.parameter_space, self.model)
+            if self.fixed_minimum is None:
+                f_min = np.min(np.append(self.model.Y, np.array(y_batch)))
+            else:
+                f_min = self.fixed_minimum
+            if self.fixed_lipschitz_constant is None:
```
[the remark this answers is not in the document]
sennendoko: Done.

emukit/core/optimization/gradient_acquisition_optimizer.py
sennendoko: Changed.

emukit/core/acquisition/integrated_acquisition.py
sennendoko: Yes. I've moved "samples" to be stored here instead, and moved the content of GPyModelWrapper.generate_hyperparameters_samples into IntegratedHyperParameterAcquisition.update_parameters, as the latter was only calling the former and was handing it a lot of attributes from IntegratedHyperParameterAcquisition - so it really belongs in that class.

emukit/bayesian_optimization/acquisitions/log_acquisition.py
```diff
@@ -17,6 +17,10 @@ def __init__(self, acquisition: Acquisition):
         """
         self.acquisition = acquisition
 
+    @property
+    def samples(self):
+        return self.acquisition.samples
+
```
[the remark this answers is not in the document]
sennendoko: Removed it without any harmful effects on tests!

emukit/core/interfaces/models.py
sennendoko: Now done.

emukit/model_wrappers/gpy_model_wrappers.py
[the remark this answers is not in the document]
sennendoko: It's now in IntegratedHyperParameterAcquisition.

emukit/bayesian_optimization/local_penalization_calculator.py
sennendoko: What's the right way to detect if the model has hyper-priors? Is it to check it's an instance of  IPriorHyperparameters?

apaleyes: Yes, that's the way we normally did it so far: `isinstance(model, IPriorHyperparameters)`

emukit/core/acquisition/integrated_acquisition.py
```diff
@@ -1,6 +1,7 @@
 from typing import Union, Callable, Tuple
 
 import numpy as np
+import GPy
```
apaleyes, line 4: I am afraid that's a big "no". Emukit core (bayes opt, exp design, quadrature, parameter space, optimization loop) cannot assume, or depend on, specific backends. Moreover, I don't see why is that needed in this case. We already have an interface that model can imlement to generate samples of hyper-parameters, and we just need to call it here.

Perhaps I am missing something here, some good reason why we cannot make a model method call. In that case we can discuss alternative approaches.

sennendoko: OK...in a narrow sense I can easily remove the dependency by removing the "import GPy" and the "if not isinstance(gpy_model, GPy.core.Model): return" - the former is just to enable the latter, and the latter is just for safety. (And I would also rename "gpy_model" to "model").

If that's not enough, I can reinstate generate_hyperparameters_samples in gpy_model_wrappers.py. I made that change simply because the call from update_parameters to generate_hyperparameters_samples involved passing a lot of attributes of IntegratedHyperParameterAcquisition into GPyModelWrapper, and the only attribute of GPyModelWrapper used inside generate_hyperparameters_samples was "model". So that piece of code seemed to me most naturally to belong in IntegratedHyperParameterAcquisition. But I don't have a good understanding of Emukit, I've just taken on the task of getting these changes merged, so I'm happy to follow guidance. Just let me know which of these options you'd like.

apaleyes: No worries, and thanks for bearing with my meticulous comments. I believe it was reasonable to have it in the wrapper, so I would suggest to keep it there, stateless as it was.

emukit/bayesian_optimization/local_penalization_calculator.py
sennendoko: Thanks - done.

emukit/core/acquisition/integrated_acquisition.py
sennendoko: Done (the latter option).

tests/emukit/multi_fidelity/test_non_linear_models.py
sennendoko: It turns out the "flatten" calls are not now needed for the tests to pass, so I've reinstated the "main" version of these tests.

emukit/model_wrappers/gpy_model_wrappers.py
```diff
@@ -108,9 +108,9 @@ def predict_covariance(self, X: np.ndarray, with_noise: bool=True) -> np.ndarray
     def get_covariance_between_points(self, X1: np.ndarray, X2: np.ndarray) -> np.ndarray:
         """
         Calculate posterior covariance between two points
-        :param X1: An array of shape n_points1 x n_dimensions. It is the first argument of the
+        :param X1: An array of shape n_points1 x n_dimensions that contains a data single point. It is the first argument of the
```
BrunoKM, line 111: This seems like a mistake, it can very well contain multiple data points.

emukit/model_wrappers/gpy_model_wrappers.py
```diff
@@ -108,9 +108,9 @@ def predict_covariance(self, X: np.ndarray, with_noise: bool=True) -> np.ndarray
...
                    posterior covariance function
-        :param X2: An array of shape n_points2 x n_dimensions. This is the second
+        :param X2: An array of shape n_points2 x n_dimensions that may contain multiple data points. This is the second
```
BrunoKM, line 113: Both X1, X2 can contain multiple data points. `that may contain multiple data points` should either appear in the doc-string for both, or none, but I feel like it's somewhat implied by the shape type-hint `n_points2 x n_dimensions`

emukit/bayesian_optimization/local_penalization_calculator.py
```diff
@@ -1,8 +1,10 @@
+from emukit.core.interfaces.models import IPriorHyperparameters
```
apaleyes, line 1:
```suggestion
from ..core.interfaces.models import IPriorHyperparameters
```

and move it a tad bit below, where the rest of relative imports are

apaleyes (commented):
I think we are nearly there. That looks good, except for one small comment below, as well as notes on docstrings made by @BrunoKM

emukit/bayesian_optimization/local_penalization_calculator.py
sennendoko: Done.

emukit/model_wrappers/gpy_model_wrappers.py
sennendoko: Done (and cleaned up a bit more too).

apaleyes (approved):
Looks good now, thanks for persisting through the review comments!

PR #383, opened 2021-06-28
Fix typos in tutorial, and add torch to requirements

sennendoko:
(1) Correcting a few typos and other minor errors in a tutorial.
(2) Adding torch to requirements.txt - this is required to run the code in emukit/examples.

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -6,3 +6,4 @@ numpy>=1.14.5
 GPy[plotting]>=1.10.0
 emcee>=2.2.1
 scipy>=1.1.0
+torch>=1.5.1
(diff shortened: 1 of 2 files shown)


requirements/requirements.txt
```diff
@@ -6,3 +6,4 @@ numpy>=1.14.5
 GPy[plotting]>=1.10.0
 emcee>=2.2.1
 scipy>=1.1.0
+torch>=1.5.1
```
apaleyes, line 9: Dependencies for examples are not added to requirements intentionally. We are lucky enough to have lots of examples in Emukit, and they carry different dependencies, non of which are required by the core lib. So the approach in Emukit is to only make users install core dependencies.

For running examples users are supposed to receive clear import errors, and manage pip install themselves.

apaleyes (commented):
All these fixes look good, thank you! Except for pytorch, on which I've commented below

sennendoko:
OK, fair enough. I've removed the torch requirement.

apaleyes (approved):

PR #384, opened 2021-07-02
moving gradients of base GP to base GP class 

mmahsereci:
*Issue #, if available:*

*Description of changes:*

Gradients:
- this is another pre-PR to #375 
- moved the gradients of the base-GP from the VanillaBQ class to the base GP where they belong. This wasn't an issue so far as VanillaBQ is basically the BaseGP just with a different API. This is still neater though, and will come in handy later.

Warpings (warping classes were introduced in PR #379 ):
- converted the base class of the warping to an abstract class. 
- the new parameter for the warping are now passed via keyword arguments. It is a bit more readable like this.

### DIFF
--- a/emukit/quadrature/methods/warped_bq_model.py
+++ b/emukit/quadrature/methods/warped_bq_model.py
@@ -74,7 +74,7 @@ class WarpedBayesianQuadratureModel(IModel, IDifferentiable):
         :returns: Predictive mean and variances of warped GP, and predictive mean and variances of base-GP in that order
                   all shapes (n_points, 1).
         """
-        raise NotImplemented
+        raise NotImplementedError
 
     def predict_base_with_full_covariance(self, X_pred: np.ndarray) -> Tuple[np.ndarray, np.ndarray, np.ndarray,
                                                                              np.ndarray]:
@@ -84,7 +84,7 @@ class WarpedBayesianQuadratureModel(IModel, IDifferentiable):
         :returns: Predictive mean and covariance of warped GP, predictive mean and covariance of base-GP in that order.
                   mean shapes both (n_points, 1) and covariance shapes both (n_points, n_points)
         """
-        raise NotImplemented
+        raise NotImplementedError
 
     def predict_with_full_covariance(self, X_pred: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
         """Compute predictive means and covariance of warped GP.
@@ -133,4 +133,4 @@ class WarpedBayesianQuadratureModel(IModel, IDifferentiable):
 
         :returns: Estimator of integral and its variance.
         """
-        raise NotImplemented
+        raise NotImplementedError
--- a/emukit/quadrature/methods/warped_bq_model.py
+++ b/emukit/quadrature/methods/warped_bq_model.py
@@ -134,12 +134,3 @@ class WarpedBayesianQuadratureModel(IModel, IDifferentiable):
         :returns: Estimator of integral and its variance.
         """
         raise NotImplemented
-
-    @staticmethod
-    def _symmetrize(A: np.ndarray) -> np.ndarray:
-        """Symmetrize a matrix.
-
-        :param A: A square matrix, shape (N, N)
-        :return: The symmetrized matrix 0.5 (A + A').
-        """
-        return 0.5 * (A + A.T)
--- a/emukit/quadrature/methods/warped_bq_model.py
+++ b/emukit/quadrature/methods/warped_bq_model.py
@@ -5,14 +5,14 @@
 import numpy as np
 from typing import Tuple, Union
 
-from ...core.interfaces.models import IModel
+from ...core.interfaces.models import IModel, IDifferentiable
 from ...quadrature.interfaces.base_gp import IBaseGaussianProcess
 from ...quadrature.kernels.bounds import BoxBounds
 from ...quadrature.kernels.integration_measures import IntegrationMeasure
 from .warpings import Warping
 
 
-class WarpedBayesianQuadratureModel(IModel):
+class WarpedBayesianQuadratureModel(IModel, IDifferentiable):
     """The general class for Bayesian quadrature (BQ) with a warped Gaussian process.
 
     Inference is performed with the warped GP, but the integral is computed on a Gaussian approximation.
@@ -111,17 +111,18 @@ class WarpedBayesianQuadratureModel(IModel):
         :param X: Observation locations, shape (n_points, input_dim)
         :param Y: Integrand observations at X, shape (n_points, 1)
         """
-        self.update_parameters(X, Y)
+        self._warping.update_parameters(**self.compute_warping_params(X, Y))
         self.base_gp.set_data(X, self._warping.inverse_transform(Y))
 
-    def update_parameters(self, X: np.ndarray, Y: np.ndarray) -> None:
-        """Update parameters of the model that are not being optimized. Override this method on case parameters
-        are data dependent.
+    def compute_warping_params(self, X: np.ndarray, Y: np.ndarray) -> dict:
+        """Compute parameters of the warping that are dependent on data, and that are not being optimized.
+        Override this method on case parameters are data dependent.
 
         :param X: Observation locations, shape (n_points, input_dim)
         :param Y: Integrand observations at X, shape (n_points, 1)
+        :returns : Dictionary containing new warping parameters. Names of parameters are the keys.
         """
-        pass
+        return {}
 
     def optimize(self) -> None:
         """Optimizes the hyperparameters of the base GP"""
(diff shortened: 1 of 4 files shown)


emukit/quadrature/methods/warpings.py
```diff
@@ -1,30 +1,34 @@
...
 
+    @abc.abstractmethod
     def transform(self, Y: np.ndarray) -> np.ndarray:
         """Transform from base-GP to integrand.
 
         :param Y: Function values of latent function, shape (n_points, 1).
         :return: Transformed values, shape (n_points, 1).
         """
-        raise NotImplemented
+        pass
```
apaleyes, line 16: It seems invalid to have a warping that doesn't do or return anything on transform. I'd say this abstract method should keep throwing an exception

emukit/quadrature/methods/warpings.py
```diff
@@ -1,30 +1,34 @@
...
 
+    @abc.abstractmethod
```
apaleyes, line 18: Same here, I think it shall be throwing NotImplemented, as before

emukit/quadrature/methods/warpings.py
```diff
@@ -1,30 +1,34 @@
...
-        raise NotImplemented
-
-    def update_parameters(self, **kwargs) -> None:
-        """Update the warping parameters."""
         pass
 
+    def update_parameters(self, **new_parameters) -> None:
+        """Update the warping parameters. The keyword arguments ``new_parameters`` contain the parameter names as
+        keys with the new values. An empty dictionary will not update any parameters."""
+        self.__dict__.update(new_parameters)
```
apaleyes, line 30: Seems kinda scary to mess with this. Is there any reason why we can't just create `_warping_parameters = {}` and update that?

emukit/quadrature/methods/warped_bq_model.py
```diff
@@ -133,3 +134,12 @@ def integrate(self) -> Tuple[float, float]:
         :returns: Estimator of integral and its variance.
         """
         raise NotImplemented
+
+    @staticmethod
+    def _symmetrize(A: np.ndarray) -> np.ndarray:
+        """Symmetrize a matrix.
```
apaleyes, line 140: Can you add a brief explanation what `to symmetrize` means?

emukit/quadrature/methods/warped_bq_model.py
```diff
@@ -133,3 +134,12 @@ def integrate(self) -> Tuple[float, float]:
         :returns: Estimator of integral and its variance.
         """
         raise NotImplemented
+
+    @staticmethod
+    def _symmetrize(A: np.ndarray) -> np.ndarray:
```
apaleyes, line 139: This will be used in later PRs, right? It isn't called anywhere in this PR

emukit/quadrature/methods/warped_bq_model.py
```diff
@@ -133,3 +134,12 @@ def integrate(self) -> Tuple[float, float]:
         :returns: Estimator of integral and its variance.
         """
         raise NotImplemented
+
+    @staticmethod
```
apaleyes, line 138: Private static method seems weird

emukit/quadrature/methods/warped_bq_model.py
```diff
@@ -111,17 +111,18 @@ def set_data(self, X: np.ndarray, Y: np.ndarray) -> None:
...
+    def compute_warping_params(self, X: np.ndarray, Y: np.ndarray) -> dict:
+        """Compute parameters of the warping that are dependent on data, and that are not being optimized.
+        Override this method on case parameters are data dependent.
 
         :param X: Observation locations, shape (n_points, input_dim)
         :param Y: Integrand observations at X, shape (n_points, 1)
+        :returns : Dictionary containing new warping parameters. Names of parameters are the keys.
         """
-        pass
+        return {}
```
apaleyes, line 125: I am a bit out of context here, can you bring me up to speed please? Is this class supposed to be used directly, or only serve as a base? If the latter, maybe consider turning it into an abstract class too.

Also, since we don't have any overrides in this PR, I assume they will be coming later, right?

apaleyes (commented):
SOme tests would be nice, perhaps for `get_predictive_gradients` and `symmetrize`

emukit/quadrature/methods/warped_bq_model.py
mmahsereci: This is the default behavior (empty dict means parameters stay the same). [Here](https://github.com/EmuKit/emukit/pull/375/files#diff-34e3d9470acc26afa739a9cf396ac75ca15dc290975d58ac165e97ed87620e53R57) is an example of it being overridden.

emukit/quadrature/methods/warped_bq_model.py
mmahsereci: I'll remove it as it is not used yet.

emukit/quadrature/methods/warped_bq_model.py
mmahsereci: see above

emukit/quadrature/methods/warpings.py
mmahsereci: well, yeah. It only is relevant really if someone feels like calling the method of the abstract super class. with `super().transform(Y)`. Then, instead of doing nothing it would throw an error. I can change it if you like.

emukit/quadrature/methods/warpings.py
mmahsereci: see above.

emukit/quadrature/methods/warped_bq_model.py
mmahsereci: The method should probably be called `symmetrize_matrix`, but I'll remove it anyways for now. It means 0.5  (A + A') where A is the matrix.

emukit/quadrature/methods/warpings.py
mmahsereci: I'll replace it with 
```
for parameter, new_value in new_parameters.items():
    setattr(self, parameter, new_value)
```

That is more explicit. In what way is it scary?

apaleyes: i was always assuming this is a read-only thingy. it probably partially generated on the fly in Python internals

mmahsereci:
Concerning tests, the tests for warpings are already merged, they are [here](https://github.com/EmuKit/emukit/blob/main/tests/emukit/quadrature/test_warpings.py).
The gradients of the GP, I just moved around, so they are being tested implicitly [here](https://github.com/EmuKit/emukit/blob/main/tests/emukit/quadrature/test_vanilla_bq.py#L131) as vanilla BQ is just a wrapper of the base GP. Is that enough, or do you think an explicit test would be better?

emukit/quadrature/methods/warpings.py
apaleyes: I do. But there is a more explicit problem there, no? If I inherit this class and don't override the method, then someone calls it and is left very perplexed

mmahsereci: I don't think you can instantiate a child class that does not override all abstract methods (it throws an exception).

emukit/quadrature/methods/warpings.py
mmahsereci: There is a test for it in the test file for warpings so I was confident it at least did the job functionality wise, but I agree it looked a bit suspicious. Couldn't find a good reason though. `setattr` seems better though.

apaleyes: Yes, setattr looks more reasonable

emukit/quadrature/methods/warpings.py
apaleyes: Oh, does it? If so, it makes no difference indeed. But! If you explicitly throw an exception, none of that dialog would have happened!

So I think the choice here is either keep `pass` and document the what you just replied above, or throw an exception.

emukit/quadrature/methods/warped_bq_model.py
apaleyes: I see. Ok, let's leave it as is then, thanks for clarification. I keep getting slightly confused about the structure of our BQ part

apaleyes (approved):
I'll leave it up to you whether you want to address the pass/exception thing, but otherwise lgtm!

emukit/quadrature/methods/warpings.py
mmahsereci: I am not sure where the confusion comes from to be honest. `pass` is the default to use with `abc.abstractmethod`. It's virtually in any abc tutorial or code you can find, so it should not be confusing in the first place (we even use it in Emukit e.g. [here](https://github.com/EmuKit/emukit/blob/main/emukit/core/acquisition/acquisition.py)). One of the main advantages about the `abc` package is in fact that you can force implementations of methods; this, one should not need to explain either. 

The example I gave where it actually matters what to write into the `abtractmethod` is quite cornercase and I hightly doubt many people know that it may have any effect to write sth other than `pass` into an abstract method.  I'll throw an exception just to be extra cautious, although it;s a bit non standard I suppose.

apaleyes: Right, then it's really me being behind on how ABC works. THanks for explanation!

PR #385, opened 2021-07-08
Validate categorical variables when used in contexts 

ndalchau:
Problem addressed
---------------------
Previously, if an `AcquisitionOptimizer` received a categorical variable, it would incorrectly attempt to check whether the value was *in-domain*. It was assuming that the encoded value was being passed (of type `np.ndarray`). If passing a `str` category, this would fail, while passing a numerical category would lead to various (mostly wrong) outcomes, without warning.

Changes
---------
- It is now possible to use categorical variables in contexts. The `ContextManager` is now solely responsible for *in-domain* checks (for continuous parameters, raising warnings upon failure) and *in-encoding* checks (for categorical parameters, raising `ValueError` upon failure).
- Also changed `any` to `Any` in context type.

Related
--------
- Issue #359 could be related to this.
- PR #381 introduced an alternative set of changes to the same problem.

### DIFF
--- a/emukit/core/optimization/acquisition_optimizer.py
+++ b/emukit/core/optimization/acquisition_optimizer.py
@@ -24,20 +24,6 @@ class AcquisitionOptimizerBase(abc.ABC):
         """
         self.space = space
 
-    def _validate_context_parameters(self, context: Dict[str, any]):
-        for context_name, context_value in context.items():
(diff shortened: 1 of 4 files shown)


apaleyes (approved):
lgtm, thanks for following through!

PR #386, opened 2021-08-05
Fix issue with creating IntegratedHyperParameterAcquisition objects

sennendoko:
This PR fixes the following problem: IntegratedHyperParameterAcquisition.update_batches was not doing what it should, because it was calling update_batches on a newly-minted Acquisition object which it immediately threw away. Then, later Acquisition objects were not having update_batches applied to them. With this PR, update_batches saves its argument as an attribute, and a new method generate_acquisition_with_batch_update creates the Acquisition object and then calls update_batches on it with the values stored in the attribute.

### DIFF
--- a/emukit/bayesian_optimization/local_penalization_calculator.py
+++ b/emukit/bayesian_optimization/local_penalization_calculator.py
@@ -58,11 +58,8 @@ class LocalPenalizationPointCalculator(CandidatePointCalculator):
         self.acquisition.update_parameters()
 
         # Initialize local penalization acquisition
-        import logging
         if isinstance(self.model, IPriorHyperparameters):
-            logging.info("local_penalization_calculator.py:63: creating IntegratedHyperParameterAcquisition instance")
             local_penalization_acquisition = IntegratedHyperParameterAcquisition(self.model, LocalPenalization)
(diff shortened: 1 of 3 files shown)


ekalosak:
Thanks for the PR and the explanation!! I have a question re: interfaces - does this somehow cause an issue for other objects expecting `Acquisition.update_batches()` to do something specific? Or have you checked that already?

apaleyes (commented):
Thanks for the fix! May I ask you to add a test for it though? I imagine a test which uses Mock as the output of `acquisition_generator` and ensures this mock is then called with expected parameters.

apaleyes:
@ekalosak i took a quick look. The only other place where `update_batches` shows up is in LP, and it is implemented correctly there.

sennendoko:
OK, test added - hope you like it!

apaleyes (approved):

apaleyes:
Hey @sennendoko ! I had to revert this change. The builds started timing out frequently on running the very first notebook. I've reproduced this issue locally, and traced it down to creation of acquisition. Reverting the change from this PR fixes the issue, so I am pretty sure it is the culprit, although not clear at all why.

Couple of examples of timed out builds:
https://github.com/EmuKit/emukit/actions/runs/1107776522
https://github.com/EmuKit/emukit/actions/runs/1106653448

If you can fix the issue, I am happy to merge these changes back in.

Issue #387, opened 2021-08-06
Issue with running multi-fidelity BO notebook.


PC-FSU:
System and package information 

OS : Windows 10, verision 1909, build : 18363.1440
Python : Python 3.7.10, (MSC v.1916 64 bit (AMD64))
Scipy : 1.7.0
Numpy : '1.21.1'
Setuptools : '52.0.0.post20210125'
GPy :  '1.10.0'
emcee : 3.1.0

I have a version of "Emukit-tutorial-multi-fidelity-bayesian-optimization.ipynb" adapted in two different .py scripts, and I am trying to replicate the result of the Forrester example.  File emukit_wrapper.py contains the functionality for Initializing the model, defining the acquisition function, and running the bayesian optimization loop.  

        `import numpy as np
        np.random.seed(12345)
        from emukit.multi_fidelity.convert_lists_to_array import convert_x_list_to_array
        from emukit.core import ParameterSpace, ContinuousParameter, InformationSourceParameter
        from emukit.multi_fidelity.models.linear_model import GPyLinearMultiFidelityModel
        import GPy
        from emukit.multi_fidelity.kernels.linear_multi_fidelity_kernel import LinearMultiFidelityKernel
        from emukit.multi_fidelity.convert_lists_to_array import convert_xy_lists_to_arrays
        from emukit.model_wrappers import GPyMultiOutputWrapper
        from GPy.models.gp_regression import GPRegression
        from emukit.bayesian_optimization.acquisitions.entropy_search import MultiInformationSourceEntropySearch
        from emukit.core.acquisition import Acquisition
        from emukit.core.loop import FixedIntervalUpdater, OuterLoop, SequentialPointCalculator
        from emukit.core.loop.loop_state import create_loop_state
        from emukit.core.optimization.multi_source_acquisition_optimizer import MultiSourceAcquisitionOptimizer
        from emukit.core.optimization import GradientAcquisitionOptimizer
        from emukit.core.loop.user_function import MultiSourceFunctionWrapper
        from scipy.optimize import minimize
        from emukit.bayesian_optimization.loops import BayesianOptimizationLoop
        from emukit.bayesian_optimization.acquisitions.entropy_search import EntropySearch
        from emukit.model_wrappers import GPyModelWrapper
        import GPy
        
        class EmukitMultifidelity():
        
            def __init__(self, fidelities, n_fidelities, costs,
                         n_optimization_restarts=5, verbose_optimization=False):
        
[...]

PR #392, opened 2021-12-09
Fix output dimensions of ES for 1 point input case

apaleyes:
*Issue #, if available:* #387

*Description of changes:* ES acquistion function turned out to be handling output shapes weirdly, which resulted in wrong shapes for the case when only one point was given as input. This PR fixes that, and adds few unit tests to validate shapes better.


mmahsereci:
hey @apaleyes , let me know whenever you want me to review anything :) I am also monitoring Emukit

apaleyes:
@mmahsereci great, will keep in mind! For now if you could go over BQ-related open issues and update them, that would be lovely. That is, if there is any chance they'll be worked on, or just closed, or at least add more context so that anyone else could pick them up.

Issue #257, opened 2019-11-01
Clean up integral tests for kernels in BQ package


mmahsereci:
make a config file from the scripts that produce the ground truth integrals

mmahsereci:
The linked PRs make it such that the MC estimators used to test the BQ integrals use the same models as the tests now. The main issue was that the two implementations (MC estimators and tests) were disconnected before. Closing.

PR #406, opened 2022-04-08
MC estimators for ground truth integrals of kernel embeddings in `quadrature` tests now use same setup as tests

mmahsereci:
*Issue #, if available:* 

#257 

*Description of changes:*

The kernel embeddings in the `quadrature` package are tested by comparing to MC estimators. The MC estimators are pre-computed to speed up tests. So far, the two implementations (tests and MC estimators) were completely separate which could lead to bugs as different models could be used. Now the script computing the MC estimators uses the same setup as the tests, hence when it is re-run it shoudl produce valid estimators and confidence intervals. It's still not super pretty but better than before.

Tests themselves are unchanged.


apaleyes:
Whoa! Welcome back, Maren!

mmahsereci:
> Whoa! Welcome back, Maren!

I was never gone ;)

PR #410, opened 2022-04-13
MC estimators for ground truth integrals of models in `quadrature` tests now use same setup as tests

mmahsereci:
*Issue #, if available:*

*Description of changes:*
The 2 outputs of the `integrate` method of the warped BQ models in the `quadrature` package are tested by comparing to MC estimators. The MC estimators are pre-computed to speed up tests. So far, the two implementations (tests and MC estimators) were completely separate which could lead to bugs as different models could be used. Now the script computing the MC estimators uses the same setup as the tests, hence when it is re-run it should produce valid estimators and confidence intervals. It's still not super pretty but better than before.

Additionally this PR parameterizes those tests which makes it easier to add new models to the tests.
Test themselves are not changed.


mmahsereci:
I'll merge this again @apaleyes . There are no functional changes in this PR. Just structuring the tests more robustly and more readable.

PR #414, opened 2022-04-27
Brownian motion `quadrature` kernel and GPy wrapper

mmahsereci:
*Issue #, if available:*

*Description of changes:*
This PR adds 
- kernel embeddings for the Brownian motion kernel w.r.t. Lebesgue measure.
- an EmuKit interface for this kernel.
- a GPy wrapper for said interface.
- the kernel is added to the tests (shape, value, gradient etc tests) by adding a fixture to the existing tests.
- some minor docstring adjustments.
- some minor fix in the `quadrature` gradient checks (the input values used would occasionally not obey the domain constraints; now they do.)
- 2 additional tests that check if the qkernels raise exceptions when provided with wrong input.

### DIFF
--- a/emukit/model_wrappers/gpy_quadrature_wrappers.py
+++ b/emukit/model_wrappers/gpy_quadrature_wrappers.py
@@ -238,7 +238,7 @@ class BrownianGPy(IBrownian):
     r"""Wrapper of the GPy Brownian motion kernel as required for some EmuKit quadrature methods.
 
     .. math::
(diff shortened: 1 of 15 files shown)


mmahsereci:
Hey @apaleyes , did we ever decided if/ how we want to test model wrappers? Some or implicitly tested via the tests/ integration tests but we don't really have dedicated tests for the module. I am OK with how it is for `quadrature` as the functional parts are being tested elsewhere and only some convenience functions are not being testes, but it makes the test coverage drop which is annoying.

PR #430, opened 2022-08-05
`ProductMatern12` kernel embedding w.r.t Lebesgue measure & GPy wrapper

mmahsereci:
*Issue #, if available:*

*Description of changes:*
This PR adds

- kernel embeddings for the Matern12 (a.k.a. Exponential) product kernel w.r.t. Lebesgue measure.
- an EmuKit interface for this kernel.
- a GPy wrapper for said interface.
- the kernel is added to the tests (shape, value, gradient etc tests) by adding a fixture to the existing tests.

### DIFF
--- a/emukit/model_wrappers/gpy_quadrature_wrappers.py
+++ b/emukit/model_wrappers/gpy_quadrature_wrappers.py
@@ -14,6 +14,7 @@ from ..quadrature.interfaces import (
     IBaseGaussianProcess,
     IBrownian,
     IProductBrownian,
+    IProductMatern12,
     IProductMatern32,
     IProductMatern52,
 )
@@ -22,6 +23,7 @@ from ..quadrature.kernels import (
     QuadratureKernel,
     QuadratureProductBrownian,
     QuadratureProductBrownianLebesgueMeasure,
+    QuadratureProductMatern12LebesgueMeasure,
     QuadratureProductMatern32LebesgueMeasure,
     QuadratureProductMatern52LebesgueMeasure,
     QuadratureRBFGaussianMeasure,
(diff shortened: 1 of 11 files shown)


tests/emukit/quadrature/test_quadrature_kernels.py
```diff
@@ -366,7 +416,7 @@ def test_qkernel_qKq(kernel_embedding, interval):
             ),
         ),
         (
-            lebesgue_embeddings_test_list[2],
+            lebesgue_embeddings_test_list[3],
```
apaleyes, line 419: Maybe these lists could be turned into dictionaries, so that such changes won't be necessary? `lebesgue_embeddings_test["qmatern32"]` also reads a bit better

emukit/quadrature/interfaces/standard_kernels.py
```diff
@@ -84,6 +84,67 @@ def dKdiag_dx(self, x: np.ndarray) -> np.ndarray:
...
+        r"""The lengthscales :math:`\lambda` of the kernel."""
+        raise NotImplementedError
+
+    @property
+    def variance(self) -> float:
+        r"""The scale :math:`\sigma^2` of the kernel."""
+        raise NotImplementedError
+
+    def _dK_dx1_1d(self, x1: np.ndarray, x2: np.ndarray, ell: float) -> np.ndarray:
+        """Unscaled gradient of 1D Matern12 where ``ell`` is the lengthscale parameter.
```
apaleyes, line 129: `ell` sounds really cryptic. WHy not just `lengthscales`?

emukit/model_wrappers/gpy_quadrature_wrappers.py
```diff
@@ -140,6 +142,120 @@ def K(self, x1: np.ndarray, x2: np.ndarray) -> np.ndarray:
...
+        return self.gpy_matern.K(x1, x2)
+
+    def _K_from_prod(self, x1: np.ndarray, x2: np.ndarray, skip: List[int] = None) -> np.ndarray:
+        """The kernel k(x1, x2) evaluated at x1 and x2 computed as product from the
+        individual 1d kernels.
+
+        :param x1: First argument of the kernel.
+        :param x2: Second argument of the kernel.
+        :param skip: Skip these dimensions if specified.
+        :returns: Kernel evaluated at x1, x2.
```
apaleyes, line 231: These need shape specs in the docs

apaleyes (commented):
Now that the build is unblocked, a couple of very minor comments

emukit/model_wrappers/gpy_quadrature_wrappers.py

tests/emukit/quadrature/test_quadrature_kernels.py
mmahsereci: good idea! done.

emukit/quadrature/interfaces/standard_kernels.py
mmahsereci: All product kernels in the `quadrature` package have methods that act on the d-dimensional space and other methods that solely act on an individual dimension of the product kernel. In order to make it clear if a method acts on all or a single dimension I used `lengthscales` and `ell` respectively (`ell` is from LaTeX's \ell command) inside the methods. I found `lengthscale` to be too similar to `lengthscales` that it confused me. 

I changed the key from "ell" to "lengthscale" in the dictionaries that define the per-dimension parameters for all the product kernels, so it's a bit more clear what `ell` is supposed to be. But let me know if it's still too confusing, then I'll change it.

apaleyes: Still very cryptic. Perhaps you could do something nice with prefixes? E.g. `full_space_lengthscale` and `single_dim_lengthscale` or something of that sort?

It actually took me a while to realise that `ell` isn't an abbreviation at all. Shame on me for sure, but also kinda highlights how confusing it can be for some

mmahsereci: Probably `lengthscales` should stay as is as it's part of the interface and it's also similar to the interface of GPy kernels. I'll change `ell` to `lengthscale` then as you originally proposed. If the variable names get too long in algebraic expressions then the code is also hard to read which may be even more confusing, especially with all the black formatting. Sounds good?

I'll do it in a quick follow up PR though if that's OK as I need to change this in all product kernels and so the changes are a bit better separated from this PR.

PR #434, opened 2022-09-02
Renaming `ell` variable to `lengthscale` in `quadrature` internal code 

mmahsereci:
*Issue #, if available:*

*Description of changes:*
See #430 for discussion on this. None of the renaming is user-facing.


mmahsereci:
Just a quick check if you think the formulas are still parsable even with somewhat lengthy variable names @apaleyes . 

All code where `ell` was used are private methods that are meant to be meddled with only by the devs. Hence readability of the algebraic expressions is important and lengthy variables can make this harder. I personally think the formulas are harder to read like this. Again, I don't have a super strong opinion and I am happy to rename, just wanted to highlight that this code is the core computation in `quadrature` and not seen by someone who simply uses the library. (Just thought I add this here because I haven't mentioned it in the other discussion).

apaleyes (approved):

Issue #461, opened 2024-04-28
Add windows build back


apaleyes:
After bumping python version in #460 build on windows started failing. Error message:

```
  UPDATING build\lib.win-amd64-cpython-310\matplotlib\_version.py
  set build\lib.win-amd64-cpython-310\matplotlib\_version.py to '3.3.4'
  running build_ext
  Extracting freetype-2.6.1.tar.gz
  Building freetype in build\freetype-2.6.1
  msbuild build\freetype-2.6.1\builds\windows\vc2010\freetype.sln /t:Clean;Build /p:Configuration=Release;Platform=x64
  error: command 'msbuild' failed: None
  [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for matplotlib
ERROR: Could not build wheels for matplotlib, which is required to install pyproject.toml-based projects
Successfully built emukit
Failed to build matplotlib
Error: Process completed with exit code 1.
```

Definitely want to keep supporting windows, so we need to figure out what's going on

apaleyes:
Thanks for the bump! Would you mind sharing the error you get?

RedLenco:
Thanks for your reply! Sure, here is the output I get after running pip install emukit:

```
pip install --no-cache-dir emukit
Collecting emukit
  Downloading emukit-0.4.10.tar.gz (170 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Requirement already satisfied: setuptools>=40.8.0 in c:\programdata\anaconda3\envs\fsms\lib\site-packages (from emukit) (68.0.0)
Requirement already satisfied: numpy>=1.14.5 in c:\programdata\anaconda3\envs\fsms\lib\site-packages (from emukit) (1.26.4)
Requirement already satisfied: GPy>=1.10.0 in c:\programdata\anaconda3\envs\fsms\lib\site-packages (from GPy[plotting]>=1.10.0->emukit) (1.13.2)
Collecting emcee>=2.2.1 (from emukit)
  Downloading emcee-3.1.6-py2.py3-none-any.whl.metadata (3.0 kB)
Requirement already satisfied: scipy>=1.1.0 in c:\programdata\anaconda3\envs\fsms\lib\site-packages (from emukit) (1.12.0)
[...]

PR #464, opened 2024-09-23
Install mpl differently, bring windows back

apaleyes:
*Issue #, if available:* #461

*Description of changes:* Instead of relying on GPy that hardcodes matplotlib version (which seems broken, see issue), let's install matplotlib separately. GPy, unfortunately, still requires matplotlib implicitly.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,7 +40,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


mmahsereci (approved):
LGTM! Thanks @apaleyes !

Just so I understand, with this workaround, some of the plotting functionality of GPy won't work correctly? (But we don't mind this). Could you also link the related GPy issue (in case there is one)?

apaleyes:
> some of the plotting functionality of GPy won't work correctly

possibly! for whatever reason GPy[plotting] fixes a very particular version of mpl that itself cannot be installed on windows

> Could you also link the related GPy issue

https://github.com/SheffieldML/GPy/issues/1034

PR #483, opened 2026-02-20
Use scipy for model free designs

apaleyes:
*Issue #, if available:* n/a

*Description of changes:* Noticed that scipy actually provides implementations of Sobol and Latin Hypercube designs. Meaning we can remove two dependencies! Plus a test to validate model free design's bounds.


mmahsereci (approved):
Thanks, Andrei! Eyeballing this a bit as I have some trouble with my python installation atm but it looks good! Glad we can get rid of this extra dependency.

PR #484, opened 2026-05-08
Adding agents md

mmahsereci:
*Issue #, if available:*

*Description of changes:*


mmahsereci:
hi @apaleyes , will still iterate on this a bit, but let me know what you think is missing or should be adjusted (perhaps some template for the PR text?). Also how interface classes starting with "I" are used could be extended...

apaleyes:
@mmahsereci looks reasonable! Overall i think agents.md is the best place to explain stuff that coding agents don't get, tricky and obscure aspects of the library. But we can only amass those with experience of using coding agents in Emukit dev. The docstring style from the other PR is a good example, and not an obvious one to think of

